Issue #100, opened 2014-05-13
The approval icon just looks wrong


timlinux:
# Problem

The approve icon is visually inconsistent - it should be gray and its size should be consistent with its neighbours.

![selection_006](https://cloud.githubusercontent.com/assets/178003/2954090/73d01314-da6b-11e3-9f39-2904b1841c0c.png)

# Proposed solution

* standardise on gray / light background as with other icons
* make the size the same as edit/remove icons
* check everywhere in the site where we use approve to make this fix globally

# Expected outcome

Visual consistency and a more professional looking site

# See also

#99

timlinux:
Please check & close

Issue #98, opened 2014-05-13
Project block should make it clear that versions listed are only two most recent


timlinux:
# Problem

When you look at the project block, it isn't clear that there may be more versions than listed:

![selection_001](https://cloud.githubusercontent.com/assets/178003/2953990/336a9e7c-da69-11e3-8541-3ae2ec014133.png)

Actually when you click into the project you see this:

![selection_002](https://cloud.githubusercontent.com/assets/178003/2953994/3dcd3186-da69-11e3-939b-e01c77c7a462.png)

# Proposed solution

Simply adding the the word 'Latest' in front of the project block versions list will be a good start.

![selection_001](https://cloud.githubusercontent.com/assets/178003/2954000/690b3b04-da69-11e3-9e88-10ebc2db6b03.png)

We could also think about making that list scrollable but it might look crap.

# Expected outcome

User should be clear that there are more versions for this project.

timlinux:
At the moment I have changed it to show last 5 versions, but essence of the ticket remains...

rischanlab:
Hi @timlinux I proposed this solution 

![screen shot 2015-10-05 at 10 48 18 am](https://cloud.githubusercontent.com/assets/2235894/10272768/0ae183a8-6b4f-11e5-947d-ed6af5e760d9.png)

if you agree you can merge my pull request 
thanks

Issue #96, opened 2014-05-08
Project page needs an 'add version' button at top of version list


timlinux:
# Problem

When a user is logged in and looking at a project page (e.g. http://changelog.linfiniti.com/inasafe/) there should be a button letting them add a new version.

![selection_001](https://cloud.githubusercontent.com/assets/178003/2913644/46bae3e8-d68f-11e3-9f7a-48fb5be679b0.png)

# Expected outcome

You should be able to directly add a new version on the project page

PR #115, opened 2015-10-05
Adding .travis.yml for routinely code testing

rischanlab:
Hi @timlinux , I gave a comment for running coverage in this commit cause I could not solve the coverage problem. When I run coverage test in my local machine I got similar error with [this] (https://github.com/WikiWatershed/model-my-watershed/issues/39) and [this](https://github.com/evonove/django-oauth-toolkit/issues/204) and I still can't find the solution. So I just cleaning projecta's code based on flake8. Now I am skipping that and continuing to fixing issues #96 #98 and #100


deployment/Makefile
```diff
@@ -211,8 +211,7 @@ mediasync:
 	@echo "------------------------------------------------------------------"
 	@echo "Syncing media files from live server"
 	@echo "------------------------------------------------------------------"
-	@rsync -av --progress kartoza3:/home/timlinux/production-sites/projecta/django_project/media media
-
+	@rsync -av --progress rischan@5.9.160.107:/home/data/changelog.inasafe.org/deployment/media media
```
timlinux: This change is bad since it will break my ability to sync media. Please revert to the original line and chat to me if it is not working for you.

deployment/docker-dev/Dockerfile-dev
```diff
@@ -0,0 +1,51 @@
+#--------- Generic stuff all our Dockerfiles should start with so we get caching ------------
```
timlinux: Newer docker compose lets you explicitly specify which docker file to use to build. So I would prefer that you simply do this:

* rename docker-prod dir to simply 'docker'
* add ``Dockerfile-dev.yml`` to that directory
* update ``docker-compose.yml`` to reference the ``Dockerfile-dev.yml`` for the dev service description
* change the ``Dockerfile-dev.yml`` to add requirements-dev.txt (which you should rename)

I will explain more on FaceTime which the motivation is behind this request if anything is not clear.

django_project/base/templates/project/detail.html
```diff
@@ -29,9 +29,20 @@
     {% endfor %}
   {% endif %}
   {% if versions %}
-    <div class="row">
-      <h3>Versions</h3>
-      <hr/>
+          <div class="page-header">
```
timlinux: @rischanlab we should be making atomic PR's. That means one PR for one github issue as fat as possible. It makes it much easier to track the changes related to an issue and also to review changes as they come in.

django_project/base/templates/project/includes/project-panel.html
```diff
@@ -70,6 +70,11 @@ <h4 class="text-muted">
             </span>
           </p>
         {% endfor %}
+          {% if project.versions.count >= 2 %}
+              <strong><span class="text-muted">Latest 2 of {{ project.versions.count }} versions</span></strong>
```
timlinux: Ok if you look in the listing you will see I send more that two now. Anytime you are hard coding numbers like this it is a sign that we might be making a wrong turn. So better to soft code 2 here to the actual threshold and counts used.

django_project/base/views/project.py
```diff
@@ -184,7 +184,8 @@ def get_queryset(self):
             return projects_qs.filter(creator=self.request.user)
 
     def get_context_data(self, **kwargs):
-        context = super(PendingProjectListView, self).get_context_data(**kwargs)
+        context = super(PendingProjectListView,
```
timlinux: Pull left applies here. Please see our coding standards doc in InaSAFE for more details.

django_project/changes/feeds/entry.py
```diff
@@ -122,7 +122,11 @@ def item_description(self, item):
         :returns: description of the Entry
         :rtype: str
         """
-        return '<p>'+item.description+'</p><p><img src="'+settings.MEDIA_URL+item.image_file.name+'"/></p>'
+
+        return '<p>' + item.description + \
```
timlinux: I know this is inherited code (i.e. you did not write it like this) but as a general pattern it is nicer to first assign the result to a variable and then return the var. The reason I prefer that is if you are debugging it is easier to evaluate what is being returned when stepping through the code.

django_project/changes/management/commands/purge_unused_images.py
```diff
@@ -32,7 +32,7 @@ def handle(self, *args, **options):
                 self.stdout.write('Moving %s' % image_path)
             image_file = File(image_path)
             entry.image_file.save(image_path, image_file, True)
-            #os.remove(image_path)
+            # os.remove(image_path)
```
timlinux: I wonder why this is commented out...maybe I was debugging and accidentally forgot to remove the # - can you check?

django_project/changes/tests/test_views.py
```diff
@@ -68,8 +69,10 @@ def test_CategoryCreate_with_login(self):
         my_response = my_client.post(reverse('category-create', kwargs={
             'project_slug': self.my_project.slug
         }), post_data)
-        self.assertRedirects(my_response, reverse('pending-category-list', kwargs={
-            'project_slug': self.my_project.slug
+        self.assertRedirects(my_response,
```
timlinux: See pull left principle in coding standards

django_project/core/settings/dev_rischan.py
```diff
@@ -0,0 +1,128 @@
+# coding=utf-8
```
timlinux: I think we can actually get rid of a lot of these dev_**user** profiles and replace with one ``dev_docker`` profile - what do you think?

django_project/core/settings/jenkins.py
```diff
@@ -1,4 +1,5 @@
-from .test import *
```
timlinux: We can get rid of this since we are shifting to travis

django_project/vota/views/ballot.py
```diff
@@ -6,7 +6,8 @@
 import logging
 from braces.views import LoginRequiredMixin, StaffuserRequiredMixin
 from django.core.urlresolvers import reverse
-from django.views.generic import DetailView, CreateView, DeleteView, UpdateView, ListView
+from django.views.generic import DetailView, \
```
timlinux: Use bracketed syntax for multiline imports please (see coding standards). i.e.

```
from django.views.generic import (
    foo, bar)
```

When all imports cannot fit in second line then do 

```
from django.views.generic import (
    foo, 
    bar,
    etc.)
```

django_project/vota/views/ballot.py
```diff
@@ -205,12 +209,10 @@ def get_form_kwargs(self):
...
         self.committee_slug = self.kwargs.get('committee_slug', None)
-        self.committee = Committee.objects.filter(project=self.project)\
-                .get(slug=self.committee_slug)
-        kwargs.update({
-            'user': self.request.user,
-            'committee': self.committee
-        })
+        self.committee = Committee.objects.filter(
+            project=self.project).get(slug=self.committee_slug)
+        kwargs.update({'user': self.request.user,
```
timlinux: pull left rules apply

timlinux:
Hi @rischanlab 

I made a bunch of comments mainly summarised as:

* make atomic PR's please
* observe ''Pull Left'  principle please
* we should observe the principle that there should be little to no application logic in templates. Rather do as much logic in the model as first priority, view as second priority and template as absolute last alternative.
* I would like to use a nicer way of arranging docker-dev - I made some comments, if it is not clear lets chat on voice
* I think there is some cruft we can clean up in terms of getting rid of user dev profiles in settings and make a simple dev_docker profile...

If you address the various items identified, then your PR looks good to me to merge.

Thanks!

rischanlab:
Hi @timlinux Could you please review my PR again? 
Thanks!

deployment/docker/Dockerfile-dev
```diff
@@ -34,7 +34,7 @@ RUN echo "export VISIBLE=now" >> /etc/profile
 RUN apt-get update -y; apt-get -y --force-yes install yui-compressor pandoc
 
 ADD REQUIREMENTS.txt /REQUIREMENTS.txt
```
timlinux: Shouldn't we be adding REQUIREMENTS-dev.txt here?

django_project/vota/tests/test_models.py
```diff
@@ -19,10 +19,10 @@ def test_Ballot_create(self):
...
 
-        #check if PK exists
+        # check if PK exists
         self.assertTrue(my_model.pk is not None)
 
-        #check if name exists
+        # check if name exists
         self.assertTrue(my_model.name is not None)
 
     def test_Ballot_read(self):
```
timlinux: test_ballot_read

timlinux:
@rischanlab LGTM

rischanlab:
Hi @timlinux I've fixed the Dockerfile-dev and for your second comment about test_ballot_read I have no idea about it.

timlinux:
Hi

My comment about test_ballot_read is that the method name is not according to PEP8 (lower case, underscore separated).

We have a philosophy that you should always leave the code in a better state than you found it. So if you touch a method you should leave it compliant to our standards - even if the mess belonged to someone else!

Regards

Tim

Issue #105, opened 2015-02-18
Wrong links to images from entry page


blazek:
If I open an entry page, for example 
    http://changelog.linfiniti.com/qgis/2.8/entry/more-responsive-browser/ and then click on the image I get server error 404 because the link is pointing to 
    /qgis/2.8/entry/<entry name>/images/entries/<hash>.<ext>
instead of 
    /media/images/entries/<hash>.<ext>

timlinux:
Note that it works fine if you go in via:

http://changelog.linfiniti.com/qgis/version/2.8/

Then click on the image.

@ismailsunni we need to get rid of all that middleware black magick and make our navitation system much simpler.

PR #116, opened 2015-10-07
Fixed Issue #100 and #105

rischanlab:
Hi @timlinux Could you please review this atomic PR?. This PR is for fixing issues #100 and #105
For the Issue #100 this is my prososed solution. 
screen shot 2015-10-07 at 10 50 53 am

For the Issue #105, I am not sure why `MEDIA_URL` does not works. I've checked in others codes such as version_list and version_details all of code use `/media/` directly instead use `MEDIA_URL`. 
So, I just change `MEDIA_URL` to `/media/` and it works.

### DIFF
--- a/django_project/base/templates/project/includes/project-panel.html
+++ b/django_project/base/templates/project/includes/project-panel.html
@@ -1,82 +1,82 @@
 {% load custom_markup %}
 {% load thumbnail %}
 <div class="panel-to-wrap col-md-4">
-  <div class="panel panel-default">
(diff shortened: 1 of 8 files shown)


timlinux:
Hi @rischanlab 

For #105 can you try the approach shown here rather:

http://stackoverflow.com/questions/3756841/django-media-url-blank

Hard coding urls in django should be considered an anti-pattern.

timlinux:
LGTM thanks

PR #117, opened 2015-10-08
Internationalization support for Projecta

rischanlab:
Hi @timlinux Could you please review my PR. Based on your request and suggestion on issue #2 I've made it. These are the examples screen when I open by english setting and Indonesian setting. (Now it supports `english`,`indonesian`,`afrikan`,`korean` ). I use `django-rosetta` for translation management that can be assess via `host/rosetta`.

screen shot 2015-10-08 at 2 12 51 pm

In Indonesian setting. 

screen shot 2015-10-08 at 2 13 09 pm

I only translate one sentence `High level software project management. Made easy.`
And about `transifex`. I've checked it, that is web application for translation collaborator. Could you give me a light, what should we do with `transifex`?

Thanks


timlinux:
LGTM

For transifex we would upload the language .po files there to be translated. Probably you will need to read some gettext materials to learn that process and then read the transifex help to see how that works.

Regards

Tim

PR #134, opened 2015-11-02
Installing Grappelli Admin to Projecta

rischanlab:
Hi @timlinux 

Please check this PR, I've Installed Grappelli admin and changed the Admin page title to `Projecta Admin`.

thanks

### DIFF
--- a/django_project/core/settings/base.py
+++ b/django_project/core/settings/base.py
@@ -128,7 +128,6 @@ INSTALLED_APPS = (
     'django.contrib.sites',
     'django.contrib.messages',
     'django.contrib.staticfiles',
-    'grappelli',
     'django.contrib.admin',
     'django.contrib.syndication',
 )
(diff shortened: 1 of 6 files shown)


django_project/core/settings/base.py
```diff
@@ -127,6 +128,7 @@
     'django.contrib.sites',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'grappelli',
```
timlinux: This should be in ``contrib.py``

rischanlab:
Hi @timlinux I've move it to contrib then grappelli does not works. I've looking some reference, some of people said that we should put `grappelli` before `django.contrib.admin`.

Then I also looking the feti project because I heard feti use grappelli. Feti also put `grappelli` like that. 
see the [base setting](https://github.com/kartoza/feti/blob/develop/django_project/core/settings/base.py) for feti project.

Thanks

timlinux:
@rischanlab if it needs to be added before you can simply do something like:

```
INSTALLED_APPS = (
    'grappelli',
) + INSTALLED_APPS
```

To insert it first into the tuple (in contrib.py).

PR #135, opened 2015-11-02
Fixes #124 | make permalink to a feature works

### LINKED ISSUE #124: Permalink to a feature is not working.
The permalink to a feature does not work. For instance ``http://changelog.linfiniti.com/qgis/version/2.10/#275`` does not scroll to the anchor.

Moreover,if we click on the anchor, the webbrowser doesn't keep the link.

rischanlab:
Hi @timlinux Now the permalink works well. Please check it. 

Thanks

### DIFF
--- a/django_project/changes/templates/version/detail-content.html
+++ b/django_project/changes/templates/version/detail-content.html
@@ -78,9 +78,9 @@
       {% if entry.approved %}
         <div class="row">
           <div class="{% if entry.image_file %}col-md-7{% else %}col-md-12{% endif %}">
-            <h3 id="feature-{{ entry.id }}">
+            <h3 id="{{ entry.id }}">
               {% if not rst_download %}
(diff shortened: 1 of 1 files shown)


django_project/changes/templates/version/detail-content.html
```diff
@@ -78,9 +78,9 @@ <h2 class="text-muted" id="category-{{ row.category.id }}">
       {% if entry.approved %}
         <div class="row">
           <div class="{% if entry.image_file %}col-md-7{% else %}col-md-12{% endif %}">
-            <h3 id="feature-{{ entry.id }}">
+            <h3 id="{{ entry.id }}">
               {% if not rst_download %}
-                <a href="#{{ entry.id }}">#</a>
+                <a href="{{ request.path }}#{{ entry.id }}">#</a>
```
timlinux: and then just add feature- in before the entry id here. I provides better semantics.

Issue #48, opened 2014-01-18
We need an app for managing sponsors


timlinux:
## Need the following fields:
- [ ] Sponsor name
- [ ] URL
- [ ] Sponsor duration
- [ ] Logo
- [ ] Contact person
- [ ] Start date
- [ ] End date
- [ ] Sponsorship agreement (as attachment)
- [ ] Level gold, silver etc.
- [ ] Sponsor email
## Views:
- [ ] A Page of all active sponsors with icons by :Large Platinum, Medium: for gold etc.
- [ ] A page with all sponsors (past and present)
## Crons
- [ ] Check daily if any sponsorships have expired and send renewal reminder

timlinux:
Sponsors should get added to the changeling entry for each release.

timlinux:
(current sponsors at the time of the release should be added only to that release)

pcav:
Of course also sponsor email should be added. A reminder (email to the sponsorship manager) sent automatically by email would be very useful. 
Thanks.

timlinux:
Thanks Paolo - I added your suggestion to the original ticket description above.

rischanlab:
Hi @timlinux Could you please check this PR #136 . This PR still on progress, but I need your comment whether my work is on right way or not. 

In this PR, admin can input the sponsors via link in navbar. 
In this case, for example admin click QGIS project then want to input the sponsors for QGIS project. 

![screen shot 2015-11-04 at 7 50 16 am](https://cloud.githubusercontent.com/assets/2235894/10926329/cdc2ba46-82c8-11e5-9643-3bb0fd147db2.png)

After that, in the page of version. When admin want to release version, he can add sponsor to that version release. 

![screen shot 2015-11-04 at 7 36 04 am](https://cloud.githubusercontent.com/assets/2235894/10926427/f4c99d7a-82c9-11e5-8989-08ab2a600743.png)

* I am using dropdown which is select all of related sponsors. I am still not sure about this, I guess one version release can have more than one sponsors, so which way is better?. Hope you can give me a comment. 


And also admin can see all of list sponsors in the page sponsor list. 

![screen shot 2015-11-04 at 8 02 56 am](https://cloud.githubusercontent.com/assets/2235894/10926495/827ddc3a-82ca-11e5-9367-92cfc885bcb8.png)


Blocker:
* I am still not able to solve `migration` so I have to make sql query when I add new model, the problem is when the table has relation with others.

andreasneumann:
[...]

PR #136, opened 2015-11-04
Adding sponsor app to projecta

rischanlab:
Refer to Issue #48 
See the comments on that issue. 

thanks


timlinux:
Hi @rischanlab 

Great to see you building this. Here are some comments:

* The Sponsors should be project specific. So users in project a should not be able to reference a sponsor in project b.
* I think we should separate amount, start and end of sponsorship from the sponsor itself so that each year when a sponsorship is renewed we do not need to re-enter everything. We also don't want to change previous sponsorship periods, so it will be better to separate them.
* Just a style note that the first line of a doctoring should be less that the 80 char line wrap, terminated by a full stop and followed by a linefeed. So:

```
This is a doc string first line that is too long because it wraps over to the next line and is not separated out from follow on content.
```

```
This is a doc string first line that is fine because it does not wrap.

This is an acceptable way to add more detail to your docstring.
```

* We should add an admin interface for sponsors.

* I didn't check carefully, but we want to list out the sponsors at the top of each version see e.g. the 2.12 release of QGIS for an example.
* I also didn't check carefully but please confirm that we have a separate table with the sponsorship levels (bronze, gold etc.) and that it supports adding an image to represent that level.

I will test it out here and add further comments as needed.

rischanlab:
Hi @timlinux 
* First, I think I did it in this PR, Just see the image below. 

screen shot 2015-11-05 at 8 19 46 pm

Sponsor must has reference to one project. When user want to add sponsor, he has to go to one project first ex: QGIS then he can add sponsor. 

* I'll think about this, I am looking the best way to do it. 
* About docstring in which file do you mean? I guess all of my docstring less than 79 but probably I miss to check in one or some files. 
* I've added the sponsor to the admin page. 
* `we want to list out the sponsors at the top of each version see e.g. the 2.12 release` I've seen the page that you mean but I cant find it, I also open the changelog.inasafe and changelog.linfiniti but I cant find it. 
* Hmm.. in this PR. for the `sponsor level` use static choice inside the model.

```
SPONSOR_CHOICES = (
    ('1', 'Platinum'),
    ('2', 'Gold'),
    ('3', 'Silver'),
    ('4', 'Bronze')
)

```
But you ask me to create different table for the sponsor level I'll do it but it takes time because I have to change `version`, `sponsor`, then add `sponsor level` because these three tables have references each others. 

Please If you have something else just tell me. 

Thanks

rischanlab:
Hi @timlinux  I was looking some ways to handle renewed sponsorship. Then this is my propose solution. I've created `sponsor renewed` table. That table only has three coulmns `sponsor` - refer to sponsor_name in sponsor table, `start_date` and `end_date`. 

I've put the `renewed sponsor button` in list of sponsors (near button `create new sponsor`).

screen shot 2015-11-06 at 2 16 08 pm

When user click that button, the renewed sponsor form will appear like this:

screen shot 2015-11-06 at 2 16 30 pm

For the sponsor field, don't worry I've filtered based on project. 

screen shot 2015-11-06 at 2 16 45 pm

Then in the list of sponsor, I'll show the all sponsor including joining table between sponsor table and renewed_sponsor table. 

How do you think about this? 
and btw about your comment `I didn't check carefully, but we want to list out the sponsors at the top of each version see e.g. the 2.12 release of QGIS for an example.` I am still not able to find what you mean. 

thanks.

timlinux:
>> and btw about your comment I didn't check carefully, but we want to list out the sponsors at the top of  each version see e.g. the 2.12 release of QGIS for an example. 

> I am still not able to find what you mean.

@rischanlab I am referring to e.g.

http://changelog.qgis.org/qgis/version/2.12.0/ - you see we have the sponsors list (compiled manually) at the top of the changelog:

screen shot 2015-11-06 at 22 58 18

What I would like is that we can do {{ project.sponsors_list }}  in the template for the version changelog and then a list like the above is rendered. Make sense?

timlinux:
I'm going to merge this - we can come back and tweak in follow up PR's - thanks @rischanlab

PR #145, opened 2015-11-08
Fix naming choices for sponsors

timlinux:


rischanlab:
Hi @timlinux I've tried your PR, it has some problems that have to fixed such as in sponsor model. I think you give a better name than mine. Anyway I'll merge it then fixed it because I want to used this one.

PR #153, opened 2015-11-13
Now we can do `migrate` on Projecta

### LINKED ISSUE #152: We can't do `migrate` on projecta

rischanlab:
Hi @timlinux After struggle for long time, I found the problem why we can't do migrate on Projecta. Fixes #152 
That's because of `AuditedModel`. 
The error that system said is 
```
File "/usr/local/lib/python2.7/site-packages/django/db/migrations/writer.py", line 465, in serialize
    "topics/migrations/#migration-serializing" % (value, get_docs_version())
ValueError: Cannot serialize: <bound method CreatorField.get_current_user of <threaded_multihost.fields.CreatorField: creator>>
There are some values Django cannot serialize into migration files.
For more, see https://docs.djangoproject.com/en/1.8/topics/migrations/#migration-serializing
```
`threaded_multihost.fields.CreatorField` is one of module that needed by `AuditedModel`.

I've tried to keep `AuditedModel` with looking the answer in google, by downgrading the django version and using south, I can do migrate with still keep `AuditedModel` but we will get error for `userena` so I think this is the best way to removed `AuditedModel`. 

I saw that this project is not update yet. [https://github.com/chrischambers/django-audited-models/](https://github.com/chrischambers/django-audited-models/) 

Now, we can do migrate. So, someone without db dump can run this projecta. 

But I still have problem, when I run `make update-migrations` that is error related with permission issue. Even I login with `make shell` then I do `python manage.py makemigrations` I got the same error. But, it works when I ssh to `devweb` and run `makemigrations` and `migrate` in there. I'll try to solved it later.

So, how do you think? are you agree to remove `AuditedModel`.


timlinux:
Hmm a tricky one - I want both migrations **and** versioning. I actually thought I had made a ticket asking for deeper use of versions by hooking into the save event of views too instead of just when used in the admin panel.

timlinux:
Ah I was confusing this with django-reversion - ok go ahead and remove the audited models.

timlinux:
LGTM

PR #162, opened 2015-12-02
fixes #147 | 404 landing page

### LINKED ISSUE #147: Landing page gives a 404

rischanlab:
Hi @timlinux 

After trial and error, its make me confused because everything is running well when debug= `True` but when we change debug=`False` (on production) it will 404 page. 

The problem is `handler404`
I just add 

```
response.status_code = 404
    return response
```

To make sure that the response is only when the status_code = 404.
and now it works well. You can open this link [http://128.199.97.119:61200/](http://128.199.97.119:61200/)

Now, it will automatically redirect on the production. 

screen shot 2015-12-02 at 1 32 44 pm

And the translation it should works too. 

screen shot 2015-12-02 at 1 34 22 pm

Thanks

### DIFF
--- a/django_project/base/views/error_views.py
+++ b/django_project/base/views/error_views.py
@@ -22,8 +22,10 @@ def custom_404(request, template_name='404.html'):
     """
     public_projects = Project.objects.filter(approved=True, private=False)
 
-    response = render_to_response(template_name, {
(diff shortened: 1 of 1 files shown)


timlinux:
LGTM

PR #163, opened 2015-12-07
Fixes #154 | removed end date

### LINKED ISSUE #154: Sponsorship end date is redundant
Since we already have the end data as a calc of start date + duration we don't need the end date field.

rischanlab:
Hi @timlinux 

I've removed sponsor end date field. 
In this PR I've changed the migration file too. 

Thanks


timlinux:
LGTM - maybe we need to do a migration too?

PR #164, opened 2015-12-08
Fixing pending items | Fixes #138

### LINKED ISSUE #138: We need to ensure that pending items shows only pending items
At the moment after you edit or create a new site, the pending items list shows all items - it should show only items awaiting moderation.

rischanlab:
Hi @timlinux 

For the pending items, as I told you before the problem is `def get_query_set` it should `def get_queryset`

I've checked that for the `category`, `version`, `sponsor` I've fixed in previous PR. 

Now, This PR is for the project, I've fixed it. 
and also I changed the redirect url, previous after user approve the project it will be redirected to the pending-items, now it will redirect to home. 

Thanks

### DIFF
--- a/django_project/base/models/project.py
+++ b/django_project/base/models/project.py
@@ -19,30 +19,30 @@ logger = logging.getLogger(__name__)
 class ApprovedProjectManager(models.Manager):
     """Custom project manager that shows only approved records."""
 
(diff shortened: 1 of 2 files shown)


timlinux:
LGTM

PR #176, opened 2016-01-04
removed migration files

rischanlab:
Hi @timlinux 

Its better if we keep the migration files or without those files? 
This PR just removed all migration files. 

If that is better to keep migration files, I'll commited the newest migration files. 

Thanks


timlinux:
I think we can remove them now and re-add them when things migrate cleanly

PR #177, opened 2016-01-14
Removed userena

rischanlab:
Hi @timlinux I've removed userena, but my big question is how we can manage the users. If we want to use another django apps. Can you recommend me? 

in this PR I also fixing the factory boy. Now we can run `test` but we still have some error in the test. 
screen shot 2016-01-14 at 4 09 39 pm

Most of errors in test is `DataError: value too long for type character varying(9)` I am not sure why its happend. 

And also in the travis the coverage error is different. I am still looking what happend. 

If you have time, you can change this PR. 

Thanks


timlinux:
LGTM

PR #178, opened 2016-01-14
Now we can run test

rischanlab:
Hi @timlinux  Now we can run test even we still have to fix many errors test, at least it runs. 

[In travis](https://travis-ci.org/rischanlab/projecta/builds/102387694)


timlinux:
LGTM

PR #184, opened 2016-01-19
Fixing sponsor view based on @timlinux comments

rischanlab:
Hi @timlinux 

This is the screenshoot of sponsor list view now. 

screen shot 2016-01-19 at 11 13 40 am

If you agree you can merge it. 
Thanks


timlinux:
Thanks @rischanlab it just needs the sponsorship level logo to go next to the sponsorship level name

[ logo ] Gold

You can also remove the underline after the sponsorship level name.

Then go ahead and merge thanks.

PR #214, opened 2016-02-11
change datetime.now to timezone.now

rischanlab:
Hi @timlinux 

I think now it solved for the tzinfo problem in [sentry](http://sentry.kartoza.com/kartoza/projecta-live/group/5848/)

In previous I was wrong to compare timezone aware datetime with naive datetime. 

Hope now it works. 

Thanks


timlinux:
Thanks @rischanlab !

PR #215, opened 2016-02-11
Sponsors view test

rischanlab:
Hi @timlinux 
In this PR, I've added some view tests for the sponsors stuff and all tests have been passed. 

btw, did you try my previous PR within your machine? 
I am bit confused for the error. 
I've tried many times with new docker containers and in my ubuntu server too, and it works well. 

The code is like this:

```
def current_sponsor(self):
        today = timezone.now()
        end = self.end_date
        if end < today:
            return False
```
end_date model 
```
end_date = models.DateField(
        _("End date"),
        help_text='End date of sponsorship period',
        default=timezone.now)
```

As my understanding, it should works. If I use `datetime` it will produce error ` can't subtract offset-naive and offset-aware datetimes`. So we have to replace `tzinfo` like in [this solution](http://stackoverflow.com/questions/4530069/python-how-to-get-a-value-of-datetime-today-that-is-timezone-aware). But that solution does not work in kartoza production even it works well in my machine. 

thanks.

### DIFF
--- a/django_project/changes/tests/model_factories.py
+++ b/django_project/changes/tests/model_factories.py
@@ -104,5 +104,5 @@ class SponsorshipPeriodF(factory.django.DjangoModelFactory):
     approved = True
     author = factory.SubFactory(UserF)
     project = factory.SubFactory('base.tests.model_factories.ProjectF')
-    sponsor_id = factory.Sequence(lambda n: n)
-    sponsorshiplevel_id = factory.Sequence(lambda n: n)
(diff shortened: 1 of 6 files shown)


timlinux:
@rischanlab it could be I am missing tz dependency on the production site - I will check. Otherwise I am not sure why it does it. I was able to replicate on my local machine too though (in devweb mode):

screen shot 2016-02-11 at 23 36 45

I am going to apply your changes and try to check if tz lib is present in production. Can you share pip freeze output with me from your django container?

PR #263, opened 2016-03-23
Fix disqus chat widget

### LINKED ISSUE #261: Disqus functionality is currently broken
There should be disqus inline chat widgets on each version page and each entry page. Currently these are not working - can we work to fix it.

dimasciput:
Fix #261


timlinux:
This should be imported ``contrib.py`` rather than ``project.py``. The ``disqus`` installed_app should also be moved to ``contrib.py``. Contrib is for third party add-ons, project is for from scratch code we created for this project.

dimasciput:
@timlinux , thanks for the information. I'll update it and merge this one.

PR #264, opened 2016-03-24
Add view to show world map of sponsors

### LINKED ISSUE #236: We need a view that shows a world map for sponsors
Requested by @aneumann

dimasciput:
Fix #236 
optimised


timlinux:
@andreasneumann would you like to make any comments bebe we merge?

timlinux:
@dimasciput this looks really great! Please merge unless Andreas has any comments

timlinux:
It might also be nice to be able to switch the map to show. Sponsoship value per country

andreasneumann:
The idea to have a sponsor world-map is good - I also thought about doing one.

I have the following feedback and ideas about it:

- - the data seems to be wrong. e.g. if you hover over Germany there are bronze sponsors that are listed with 1500€ instead of 500€, Perhaps it accumulates all past sponsorships instead of just listing the current one. I think it would be better to show the current situation than the accumulated past

- - several silver sponsors are missing (e.g. in Germany, Austria and Switzerland), but some are listed

- - in cartography it is very much discouraged to use absolute numbers in fills of polygons. You would have to put into relation with either population or BIP. Otherwise you can't compare countries properly. It is quite clear that a country with higher BIP or population has on average a higher number of donations and sponsorships. Alternatively, one could display absolute numbers with scaled symbols or charts. I have data an population and BIP extracted from Wikipedia which would allow to show relative numbers instead of absolute numbers

- - it would be good to also list donations per country and targeted donations (e.g. contributions to our bug fixing effort). I have data for both. Then we could do a pie-chart per country.

PR #269, opened 2016-03-28
Sort and group sponsors by sponsorship level and add select option element

### LINKED ISSUE #266: Sponsors map hover list should be sorted and grouped by sponsorship level
See also #264

### LINKED ISSUE #267: Sponsors map should be switchable between sponsor count per country and income per country
See also #264

dimasciput:
Fix #266 Fix #267 
projecta-optimized


timlinux:
This looks great - can you resolve merge conflicts and then we can apply?

Issue #235, opened 2016-02-19
We need to support invoice generation for sponsors


timlinux:
From @andreasneumann 

> Could we add a function to generate a sponsorship invoice?

andreasneumann:
I'll send an invoice example by mail. Contains confidential data - so it shouldn't be publically on github.

I can prepare a "neutral" example though.

timlinux:
We can use something like this to do the report generation

https://github.com/simonluijk/django-invoice

andreasneumann:
Hi Tim,
https://docs.google.com/document/d/1RMujVZ8tXEHzKYjrXVQbxqE34M9VTyuwCGqntsmJdww/edit is a template. Feel free to adapt. Content in [ ] should be replaced by dynamic content. In case of address, since this is quite different from country to country, I would agree, we could have a free-form address like you already implemented it in Projecta last weekend.

timlinux:
See #272

PR #272, opened 2016-03-30
Sponsor invoice view

dimasciput:
This PR is part of #235, I'm using easy-pdf to render pdf from html.
There are still some dummy data, like invoice number, bank account, etc.
The result is look like this: 
pdf-invoice

@timlinux, if you can give a feedback that would be great. Thanks.


timlinux:
Hi @dimasciput (cc @andreasneumann) 

That is looking pretty good. Here are a couple of comments that would need to be addressed. Maybe @andreasneumann has more:

1. Can you add two fields to the project model: sponsorship_contact_name (for holding full name of sponsorship contact person e.g. Andreas Neumann) and sponsorship_contact_email (for email of sponsorship contact)
1. Then make the pdf name at the bottom (where currently you have Jim Newton) use the sponsorship_contact_name
1. The creating of sponsor invoices should be restricted to project admins. We need to create a new role for this and then hide the sponsorship management options for people who do not belong to that role. We also need to make an association between the sponsorship role and the project so that a sponsorship role in one project does not imply that role in another project.

I think there are probably more nuances we should take care of. I think we should spin up a test instance on kartoza5 so that @andreasneumann can give more detailed comment based on actual usage.

andreasneumann:
The idea to have a sponsor world-map is good - I also thought about doing one.

I have the following feedback and ideas about it:

- the data seems to be wrong. e.g. if you hover over Germany there are bronze sponsors that are listed with 1500€ instead of 500€, Perhaps it accumulates all past sponsorships instead of just listing the current one

- only bronze sponsors are listed, no silver sponsors (we have several, e.g. in Germany, Austria and Switzerland)

- in cartography it is very much discouraged to use absolute numbers in fills of polygons. You would have to put into relation with either population or BIP. Otherwise you can't compare countries properly. It is quite clear that a country with higher BIP or population has on average a higher number of donations and sponsorships. Alternatively, one could display absolute numbers with scaled symbols or charts. I have data an population and BIP extracted from Wikipedia which would allow to show relative numbers instead of absolute numbers

- it would be good to also list donations per country and targeted donations (e.g. contributions to our bug fixing effort). I have data for both. Then we could do a pie-chart per country.

andreasneumann:
hm - seems like in Ireland a silver sponsor is shown, but not the ones in Austria, Switzerland and Germany - maybe a date-range problem?

andreasneumann:
Looks good already - thanks a lot for your work!

I think it is totally ok if parts of the invoice is static (like the header, bank information, etc.)

- In the header you need to replace the name, org name and function
qgis_invoice_header

- Officially, our organization is called "QGIS.ORG" - in the legal entity. So we should use that name in the header.

- The invoice-nr needs to be editable. I usually use a combination of year, month and running number, E.g. QGIS-Int 2016-03/07 means it is the 7th invoice in March 16. Would be cool if, when generating the invoice, I could edit that invoice number.

- the organization name of the bank account is still "QGIS usergroup Switzerland", until we finalize the paperwork of the QGIS.ORG bank account application

Thans again!

timlinux:
Can you address the comments from @andreasneumann so we can close this @dimasciput ?

timlinux:
Hi @dimasciput can you pick up on this again? @andreasneumann is going to add a few more comments - it would be good to get this merged after making the needed tweaks requested by Andreas. Please make sure everything is generic so that it works equally well for other projects.

andreasneumann:
The organization is now officially "QGIS.ORG"

The bank account information is:

Account name: QGIS.ORG
Address:      Boeschacherstrasse 10a, CH-8624 Gruet
BIC/SWIFT:    POFICHBEXXX 
IBAN:         CH09 0900 0000 9146 3839 8
Reference:    Donation QGIS
Currency:     EUR
Bank name:    PostFinance AG 
Bank address: Mingerstrasse 20, 3030 Berne, Switzerland

Other than that the above information is still valid, that we need the invoice-nr and amount to be editable/overridable.

timlinux:
Closing as superceded by #661

PR #280, opened 2016-04-03
New Nav menu for the Projecta

rischanlab:
New menu view

screen shot 2016-04-03 at 11 33 24 pm


timlinux:
Ah that looks so much better thanks. Can you resolve the conflicts please?

Issue #295, opened 2016-06-10
Hide all Pending and management menus from non staff users


timlinux:
Currently anonymous and logged in users without staff privileges can see lots of stuff they shouldn't. 

<img width="374" alt="screen shot 2016-06-11 at 00 48 05" src="https://cloud.githubusercontent.com/assets/178003/15980770/3d126d22-2f6e-11e6-87df-fe421bf551fa.png">
- [x] Don't show pending approval to non logged in users.
- [ ] Non staff users should see only entries pending approval menu item if they themselves have authored entries which are pending approval

<img width="183" alt="screen shot 2016-06-11 at 00 47 55" src="https://cloud.githubusercontent.com/assets/178003/15980772/3d7252c8-2f6e-11e6-8731-5dc5af77fa89.png">
- [x] Sponsor management should only be available for staff users
- [x] All users should be able to view the sponsors list and maps

<img width="1399" alt="screen shot 2016-06-11 at 00 47 46" src="https://cloud.githubusercontent.com/assets/178003/15980773/3d783418-2f6e-11e6-96d5-125107f6c759.png">
- [x] Changelogs menu should be staff only
- [x] Committees menu should move just to the right of the project name, left of the changelogs. Probably nicer to call it teams.

<img width="223" alt="screen shot 2016-06-11 at 00 47 01" src="https://cloud.githubusercontent.com/assets/178003/15980771/3d6c6eb2-2f6e-11e6-99ff-8d4693bc863b.png">
- [x] "Manage *" sponsor menu entries should be staff only
- [ ] We need to see the sponsor list and map menu entries (and they should appear first in the menu for staff)

<img width="1405" alt="screen shot 2016-06-11 at 00 46 47" src="https://cloud.githubusercontent.com/assets/178003/15980774/3d7ef758-2f6e-11e6-86fd-91b3a60524ea.png">
- [x] Changelogs menu should be hidden for non staff

timlinux:
BTW if I assign you both to a ticket, please collaborate to decide who will do it!

timlinux:
Can you guys review this and mark off closed items? Thanks!

PR #299, opened 2016-06-16
Hide pending and management menu from non staff 

dimasciput:
Issue #295

- [x] Don't show pending approval to non logged in users.
- [x] All users should be able to view the sponsors list and maps
- [x] Changelogs menu should be staff only
- [x] Committees menu should move just to the right of the project name, left of the changelogs. Probably nicer to call it teams.
- [x] Sponsor management should only be available for staff users
- [ ]  Non staff users should see only entries pending approval menu item if they themselves have authored entries which are pending approval

### DIFF
--- a/deployment/docker/REQUIREMENTS-dev.txt
+++ b/deployment/docker/REQUIREMENTS-dev.txt
@@ -2,7 +2,7 @@ django-nose
 coverage
 pep8>=1.5.7,<1.6
 pylint
-flake8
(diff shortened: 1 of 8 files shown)


timlinux:
LGTM so far - @rischanlab are you going to take care of the last item?

dimasciput:
@rischanlab will continue the last item on another PR, I'll merge this one.

Issue #298, opened 2016-06-15
Not able to add entries to a new project


timlinux:
Try this:

Create a new project
Create a new version
Create a new entry
Try to save it

You get a 500 error. Please investigate and fix

rischanlab:
Hi @timlinux I've checked in my local and in live server too and everything working fine. 
Please see the screenshoot below:

Create new project:

![screen shot 2016-06-20 at 9 50 09 am](https://cloud.githubusercontent.com/assets/2235894/16182121/bd06c676-36cc-11e6-9ab9-266b63cd68d2.png)

Result of new project creation.  
![screen shot 2016-06-20 at 9 50 26 am](https://cloud.githubusercontent.com/assets/2235894/16182126/cdfbd318-36cc-11e6-9492-32a40cb0c7f7.png)


The problem only in the version, when there is no version. We forget to show the button to create new version. 

![screen shot 2016-06-20 at 9 51 57 am](https://cloud.githubusercontent.com/assets/2235894/16182142/edddbb7e-36cc-11e6-9630-87efadc7156a.png)

For the category is fine. 
![screen shot 2016-06-20 at 9 54 27 am](https://cloud.githubusercontent.com/assets/2235894/16182155/04c9da7a-36cd-11e6-8b00-6d60348809bd.png)


Anyway I removed that test project again, because its on the live. 


Thanks.

timlinux:
Ok thanks. I tested on my local and it is working well here too.

PR #302, opened 2016-06-20
Added "add new version" when version is empty

rischanlab:
Hi @timlinux  this PR based on Issue #298 

Now we are able to add version when the version is empty. 

screen shot 2016-06-20 at 11 55 15 am


timlinux:
Thanks @rischanlab !

Issue #30, opened 2014-01-08
We need to implement sortable behaviour for entries and categories


timlinux:
Project manager user should be able to drag and drop the order of version entries and categories

timlinux:
This is high priority - we should be able to control exactly what order categories are rendered in, and exactly what order the entries within a category are rendered in.

rischanlab:
Hi @timlinux currently the list of version entries are mixing of all categories. so, you mean, I can put category drowpdown in the top of pages. So the admin can select which category that he want, then the list will show according to selected category. is that correct? 

thanks.

rischanlab:
* add sort number in entries model
* add button for sorting in the entries and catagories page.s

timlinux:
Its more is two buttons next to edit buttons that let you shift the entry up and down in the sort order

timlinux:
@dimasciput can you make this a priority for 1.6? I'd like to be able to sort the categories for the coming QGIS release. @nyalldawson feel free to give hints of your expectations. My expectation is that the view here : http://changelog.qgis.org/en/qgis/category/list/ would support user drag and drop to order the categories. Then any categories shown in a version would follow that order.

timlinux:
Looking at how github has implemented draggable pattern for subtasks such as mine above would be good to copy....

timlinux:
@rischanlab can you try to get this done in time for the upcoming QGIS release in about 2 weeks?

dimasciput:
@timlinux working on it

timlinux:
@dimasciput awesome thanks!

timlinux:
@ann26 can you pick this up from @dimasciput as your first priorty please? Check if he has any existing code that you can continue from first ok?

nyalldawson:
@timlinux isn't this already in?

timlinux:
@nyalldawson It is for categories, but not for entries within a category

![screencapture](https://cloud.githubusercontent.com/assets/178003/23643392/281852d6-030a-11e7-9074-cb5c147385b2.gif)

timlinux:
@Gustry can you pick up this issue too? It would be really nice to have for QGIS 3.0 release where we have a **lot** of entries and it where we could tell the story nicely by making them sortable...

PR #314, opened 2016-06-23
Sort/Order function for categories

dimasciput:
For issue #30
optimised


timlinux:
That is awesome @dimasciput ! I'm going to merge this, but could I ask you to do a follow up PR with:

* add some tests 
* and perhaps a little text at the top of the page indicating that the sort order is automatically saved (perhaps have an indicator to the right of the order categories title showing the user that the current state was saved successfully)
* Put a button next to the order categories title which will let the user go back to the categories management view.

Thanks!

timlinux:
@nyalldawson note for you that categories can now be sorted in the changelog. I have made an initial order - please feel free to tweak if needed. Individual entry sorting within categories will come in another release in the future.

nyalldawson:
@timlinux Thank you! I've made some more tweaks to the ordering to make them flow better. much appreciated!

PR #318, opened 2016-06-24
Additional function for order categories

dimasciput:
- Add some tests
- Add a little text at the top of the page indicating that the sort order is automatically saved 
- Put a button next to the order categories title which will let the user go back to the categories management view.


timlinux:
Thanks @dimasciput

Issue #310, opened 2016-06-22
Regression: Pending entries menu not available


timlinux:
With the recent pending entries cleanup we have a regression: If I edit or add an entry, it no longer appears in the pending menu (whole menu is missing) so I cannot approve entries from others.

timlinux:
@dimasciput just bumping this as the most important issue at the moment because currently we can't review updates / additions to entries...

timlinux:
[timlinux@mbp ~/dev/python/projecta ((d35644c...)|BISECTING)]$ git bisect good
bc4dc6987d840fd894044cb649a2e87a27f0688b is the first bad commit
commit bc4dc6987d840fd894044cb649a2e87a27f0688b
Author: dimasciput <dimas.ciputra@gmail.com>
Date:   Mon Jun 20 11:41:21 2016 +0700

    Pending menu only shows items that have pending approvals

:040000 040000 9f44aab5a6831c7e0e286527950d692f99ae5f14 74474c8205278f2c90a0c00df06a3cace053dff9 M	django_project

timlinux:
![screen shot 2016-07-01 at 11 26 52 pm](https://cloud.githubusercontent.com/assets/178003/16534893/89695908-3fe3-11e6-8032-608122c05eea.png)

As per the screen above, the items in the box should not be in the pending menu (they were moved to the sponsors menu), the rest should be there if there are indeed pending items for that model. Then with you path in #319 we would have one extra menu entry named 'All pending items'

PR #319, opened 2016-06-24
Add all pending entry view

dimasciput:
- Added menu to open all pending entry in the project detail page
- Added test

### DIFF
--- a/django_project/changes/templates/entry/all-pending-list.html
+++ b/django_project/changes/templates/entry/all-pending-list.html
@@ -1,0 +1,48 @@
+{% extends "project_base.html" %}
+{% load custom_markup %}
+{% load thumbnail %}
+{% load embed_video_tags %}
+{% load disqus_tags %}
+
+{% block title %}Entries - {{ block.super }}{% endblock %}
+
+{% block extra_head %}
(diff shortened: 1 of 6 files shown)


dimasciput:
Fix #310

timlinux:
Thanks for your patch @dimasciput - it still only partially fixes the problem. It should be that when I am on a sub page for a project, I should see a menu on the right if there are pending items of any sort (entry, version etc.) With your change I see it again here:

screen shot 2016-07-01 at 12 35 38 am

The pending menu was lost on the version view with a recent update - this is quite critical as there is now no easy way to look at the queue of new items that need to be reviewed. Could you make sure that menu shows in the version view too please?

CC @nyalldawson FYI

timlinux:
Also the new views you added are nice - but they do something different to the original. We should rename your new menu items to "All pending versions", "All pending entries" to clearly separate these new menu entries from the old one which showed just the pending entries missing for the version that the user is busy working on.

timlinux:
Here is an example of the view we are trying to get to with the missing menu:

http://changelog.qgis.org/en/qgis/2.16.0/pending-entry/list/

PR #326, opened 2016-07-01
Fix version indent when there is no icon

### LINKED ISSUE #325: Fix version indenting if there is a missing icon
screen shot 2016-07-01 at 12 38 56 am

Please pull the text for version numbers over to the right if there is no icon so they all line up.

dimasciput:
Fix #325 

screenshot from 2016-07-01 12 06 53


timlinux:
Thanks!

PR #335, opened 2016-07-05
Pending items broken

rischanlab:
Hi @timlinux 

There are two missing `</a>` so it makes page broken. 
screen shot 2016-07-05 at 1 17 54 pm

And also, the head of table now is hidden when sponsorship period is empty. 

Previously:
screen shot 2016-07-05 at 1 20 44 pm

Thanks.


timlinux:
Thanks @rischanlab

Issue #239, opened 2016-02-20
We need to clean up navigation


timlinux:
After creating an entry, you move to the unapproved list. On the menu bar the link to navigate back to the release entries takes you to a different view than the one you started out on.

e.g. Instead of taking you to http://changelog.inasafe.org/en/qgis/2.14.0/entry/list/, it should simply take you to http://changelog.inasafe.org/en/qgis/version/2.14.0/

timlinux:
The menu implementation has changed since writing this ticket. We still need a good way to navigate from the a submitted entry back to the original version page.

PR #345, opened 2016-07-14
Add back button on entries page

dimasciput:
To address this issue https://github.com/kartoza/projecta/issues/239 I added back button on entries page. What do you think?

screenshot from 2016-07-14 10 08 13


timlinux:
That looks nice - would be good to add it in the entry detail view too!

PR #346, opened 2016-07-14
Display thumbnail image in modal window

### LINKED ISSUE #343: Display thumbnails in a modal window when we click on fullscreen
We can see a lot of GIF in the QGIS changelog. These thumbnails are too small to see so I have to click on the button to see it fullscreen. For now, it redirects to the GIF url like http://changelog.qgis.org/media/images/entries/53f72a9cf1bf32d73eb5174c37e54c60002b9707.gif
The user needs to use the "previous" button in the web browser to come back to the changelog.

It would be better to implement a javascript modal window to show the GIF and to stay on the URL http://changelog.qgis.org/en/qgis/version/2.16.0/

dimasciput:
Fix #343 

screenshot from 2016-07-14 14 19 12

### DIFF
--- a/django_project/changes/static/js/entry.js
+++ b/django_project/changes/static/js/entry.js
@@ -1,0 +1,17 @@
+/**
+ * Created by Dimas Ciputra <dimas@kartoza.com> on 15/07/16.
+ */
+
+$(".pop-image").on("click", function() {
+    $('#imagepreview').attr('src', $(this).children().attr('id'));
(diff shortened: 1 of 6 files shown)


timlinux:
Can you put the javascript in a js file rather? Its cleaner than nesting it straight in the template / html ....

dimasciput:
@timlinux updated

timlinux:
Looks good to me! Please merge when conflicts are resolved.

Issue #340, opened 2016-07-07
rst export: please do not break lines on 70(?) chars length


rduivenvoorde:
This breaks several rst constructs like links and comment/quotes (the ones with | in it)

rischanlab:
Hi @timlinux we are using pypandoc for converting and I guess that is default from that library for rst converter they break every lines. I have tried to convert to another format and for the rst that is breaking line default. 

So, should I looking another package, or create a function for rst converter which will not breaking the lines? 

Thanks.

rduivenvoorde:
@rischanlab I'm actually waiting for a 'go' on taking the rst and publishing it on the qgis website.

So unless you think to fix this in a few hours (and not generate other issues), I would not search for a solution.

I see pandoc has an option --no-wrap, can we maybe use that?

rischanlab:
Thanks @rduivenvoorde I am working on this :)

rischanlab:
Hi @rduivenvoorde thanks for your comment. It is working https://github.com/kartoza/projecta/pull/348

dimasciput:
Fixed by #348

PR #348, opened 2016-07-15
Export rst without break lines

rischanlab:
Hi @rduivenvoorde 
Thanks for your comment https://github.com/kartoza/projecta/issues/340. It seems now is without break line. 
screen shot 2016-07-15 at 1 56 55 pm


dimasciput:
LGTM

PR #355, opened 2016-07-18
fixing vota model

meomancer:
fixing vota model bug that will create new migrations every "make update-migrations" executed

_note_:
django_project/changes/migrations/0006_auto_20160718_0804.py is added too because it seems that when SponsorshipPeriod meta model was changed but not migrated yet


cchristelis:
LGMT

Issue #347, opened 2016-07-14
Category management is broken


timlinux:
* Creating a new category gives 500 when approving
* Creating a new category gives 500 when deleting a pending category
* Duplicate categories are allowed for a project (should be unique together)
* When editing an entry, category list in entry form is not filtered to the project of the entry
* Category list should be not paginated

dimasciput:
- [x] Creating a new category gives 500 when approving
- [x] Creating a new category gives 500 when deleting a pending category
- [x] Duplicate categories are allowed for a project (should be unique together)
- [x] When editing an entry, category list in entry form is not filtered to the project of the entry
- [x] Category list should be not paginated

PR #356, opened 2016-07-18
Fix category management

dimasciput:
Issue #347 
### **Fixed:**
- Fix duplicate category name
screenshot from 2016-07-18 13 56 54

- Fix category list in entry form is not filtered to the project
- Remove pagination in category list

### **Not fixed:** 
- Creating a new category gives 500 when approving ( **_Can't reproduce this error_** )
- Creating a new category gives 500 when deleting a pending category ( **_Can't reproduce this error_** )


timlinux:
Looks good to me! thanks

PR #359, opened 2016-07-18
fixing bug on pending version acceptance

meomancer:
here is the changes for this PR:
- version cannot be created for pending project
- pending version just be shown for logged in user (for the future, just for project administrator and collaborator, related with : https://github.com/kartoza/projecta/issues/349)
- pending version in project panel will redirect too pending version list view
- when changing pending project, it will be redirect to pending project list instead detail that not found
- make cannot access pending version listview for pending project 

_authenticated_
screenshot from 2016-07-18 16 21 59

__not authenticated_
screenshot from 2016-07-18 16 22 09

this fix #357

PR #360, opened 2016-07-19
Revert "fixing bug for image that go to detail for unapproved project"

meomancer:
Reverts kartoza/projecta#358


meomancer:
i think we don't need to revert the failed travis PR, because i just changes html and already pull latest code in this PR : https://github.com/kartoza/projecta/pull/359
the error is because i merge it before travis is done

i am sorry, i will not let this happen again

PR #363, opened 2016-07-21
Update sponsor detail view

### LINKED ISSUE #362: Projecta reveals personal data without login
I noticed that one can view personal data on Projecta without logging in - see f.e. http://changelog.qgis.org/en/qgis/sponsor/uhcnwf/ which reveals person names and E-Mail addresses without a Login. I think this should be avoided and should only be viewed after  #login.

dimasciput:
- Hide personal data for unauthenticated user ( Fix #362 )
- Update sponsor detail layout

**Sponsor view for authenticated user**
authenticated

**Sponsor view for unauthenticated user**
not authenticated

### DIFF
--- a/django_project/changes/templates/sponsor/detail.html
+++ b/django_project/changes/templates/sponsor/detail.html
@@ -14,59 +14,99 @@
 
     <div class="row">
         <div class=".col-md-4">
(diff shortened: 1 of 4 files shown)


timlinux:
Looks good to me!

Issue #192, opened 2016-01-28
The 'report an issue' link is broken


timlinux:
It should be submitting an issue to github. Either fix it or just make it redirect to https://github.com/kartoza/projecta/issues/new

PR #365, opened 2016-07-21
Fixing #192

cchristelis:
Submit github issue

### DIFF
--- a/django_project/github_issue/static/js/github-issue.js
+++ b/django_project/github_issue/static/js/github-issue.js
@@ -1,5 +1,5 @@
 //noinspection JSUnusedGlobalSymbols
-function submitIssue() {
+window.submitIssue = function () {
   $("#issue-submit-button").attr("disabled", "disabled");
   $.post("/github-issue", {
     'title': $("#issue-title").val(), 'desc': $("#issue-description").val() })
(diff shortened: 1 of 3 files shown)


cchristelis:
@dimasciput, will you please verify this.

dimasciput:
@cchristelis yes the javascript works.

timlinux:
Did someone confirm we can actually submit an issue with this fix in place?

django_project/github_issue/static/js/github-issue.js
```diff
@@ -1,5 +1,5 @@
 //noinspection JSUnusedGlobalSymbols
-function submitIssue() {
+window.submitIssue = function () {
   $("#issue-submit-button").attr("disabled", "disabled");
   $.post("/github-issue", {
```
dimasciput: This url won't work, you need to remove the slash, so it will be `$.post("github-issue")`.

dimasciput:
@timlinux just need minor fix ( see my comment above ), after it fixed everything works fine. And you will need to put GITHUB_USER, GITHUB_PASSWORD, GITHUB_URL in settings file.

timlinux:
@dimasciput ok cool. I tried to disable LGTM again but it seems to be stuck :-P

timlinux:
LGTM

timlinux:
BTW @dimasciput it would be good to update the readme with instructions for issue submitting (if it is not already there). I guess it should go in secret.py....nice to have those in a template secret.py.templ (again I didn't check if it is already there).

PR #361, opened 2016-07-19
add administrator and collaborator

meomancer:
i am still struggling with when dbrestore, make migrate will break with error
`django.db.utils.ProgrammingError: there is no unique constraint matching given keys for referenced table "base_project"`

i tested it with clean db

and i don't know why travis is not building in this PR (although i already added test cases)
this partial fix #349

PR #374, opened 2016-07-26
Get repositories from github account

### LINKED ISSUE #350: Get repositories from github account
# Problem

We need to get projects based on a user's github account.

# Proposed solution

Read project information from github account and populate.

dimasciput:
Fix #350

## **User have github account connected**

out-1

Selected projects will be added to pending list
screenshot from 2016-07-26 17 11 13

## **User doesn't have github account connected**

screenshot from 2016-07-26 17 11 35

screenshot from 2016-07-26 17 11 49


timlinux:
Do you have something in place to prevent duplication if user is member in the same repos as another user?

Maybe we need to also get org model implemented first before implementing this? See https://github.com/kartoza/projecta/pull/361#issuecomment-235039246

timlinux:
Also maybe we should have a UI like this:

https://lgtm.co/kartoza

qgis

timlinux:
Ok I had a play with it in my local dev environment. It's looking nice! - we definitely need the ability to see different orgs that you are a member of though - at the moment if shows only my personal repos and forks which is not where my most interesting stuff is...

I'm going to merge this for now and then ask you to do a follow up PR with org support ok?

timlinux:
screen shot 2016-07-28 at 1 05 18 am

dimasciput:
> Ok I had a play with it in my local dev environment. It's looking nice! - we definitely need the ability to see different orgs that you are a member of though - at the moment if shows only my personal repos and forks which is not where my most interesting stuff is...

Sure, I'm going to try to get repos from different orgs.

PR #375, opened 2016-07-27
Use chosen library to make autocomplete dropdown

### LINKED ISSUE #367: When creating a team, use a text box with ajax completion rather than a select list
When there are many members in an organisation it is impossible to manage and quickly select a person. We should ideally also show their name, not their user name.

screen shot 2016-07-22 at 12 11 19 am

We should do the same anywhere else a user name is needed in a form.

dimasciput:
Fix #367

I'm using chosen library from harvest. 
https://github.com/harvesthq/chosen

##  Screenshots
**Single selection**
screenshot from 2016-07-27 15 54 49

**Multiple selection**
screenshot from 2016-07-27 16 04 46


timlinux:
That looks great thanks @dimasciput ! LGTM

PR #378, opened 2016-07-28
Fix undefined permalink function due to js compression

dimasciput:


timlinux:
Looks like this failed testing...can you ensure that travis passes before merging?

PR #381, opened 2016-07-29
Shows permalink button to non logged users

### LINKED ISSUE #379: Permalink button only shows for logged in user
We should make it appear for non logged in users too....

dimasciput:
Fix #379


cchristelis:
LGTM

Issue #384, opened 2016-09-27
Cannot add new sponsor


andreasneumann:
I tried adding new sponsors to the QGIS sponsors list, but got the following error:

500
Ooops. Something broke....

The powers that be have been informed.
In the meantime, please form an orderly queue and head to the home page.

If you need assistance, you may reference this error as 1f96dfdab8c642789dcb40867a97e90d.

I used foreign characters, f.e. "ü" and "é" in the contact names, if that matters.

Gustry:
Can we close this issue ?

andreasneumann:
yes - no issues currently.

PR #391, opened 2017-03-09
add unidecode to support unique characters from user input

ann26:
for issue #384 
I've add unidecode to decode the special characters into normal characters for the url. Therefore, when user inputs special characters, the web won't generate errors.

### DIFF
--- a/deployment/docker/REQUIREMENTS.txt
+++ b/deployment/docker/REQUIREMENTS.txt
@@ -31,3 +31,5 @@ django-embed-video==1.0.0
 django-grappelli==2.7.3
 pytz
 django-allauth==0.25.2
+#support special characters
+Unidecode==0.4.20
--- a/deployment/docker/REQUIREMENTS.txt
+++ b/deployment/docker/REQUIREMENTS.txt
@@ -26,10 +26,10 @@ django-disqus==0.5
 awesome-slugify==1.4
 #translation management tool
 django-rosetta==0.7.0
-#youtube embed
+# youtube embed
 django-embed-video==1.0.0
 django-grappelli==2.7.3
 pytz
 django-allauth==0.25.2
-#support special characters
+# support special characters
 Unidecode==0.4.20
--- a/django_project/base/models/project.py
+++ b/django_project/base/models/project.py
@@ -4,6 +4,7 @@ import os
 import logging
 import string
 import re
+import unidecode
 from django.core.urlresolvers import reverse
 from django.utils.text import slugify
 from django.conf.global_settings import MEDIA_ROOT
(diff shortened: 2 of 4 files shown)


ismailsunni:
Hi @ann26 
CMIIW, I think this `unidecode` is not python default module. Therefore, you need to add it in the https://github.com/kartoza/projecta/blob/develop/deployment/docker/REQUIREMENTS.txt so that the other developer will also have this requirement.

django_project/base/models/project.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import string
 import re
+import unidecode
```
ismailsunni: You don't need to import this one if you will only use unidecode and import it also below

django_project/changes/models/sponsor.py
```diff
@@ -10,6 +11,8 @@
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth.models import User
 from django_countries.fields import CountryField
+from unidecode import unidecode
```
ismailsunni: Same comment as above

django_project/base/models/project.py
```diff
@@ -128,7 +130,7 @@ def save(self, *args, **kwargs):
         if not self.pk:
             words = self.name.split()
             filtered_words = [t for t in words if t.lower() not in STOP_WORDS]
-            new_list = unicode(' '.join(filtered_words))
+            new_list = unidecode(' '.join(filtered_words))
```
ismailsunni: If it's not obvious what it does, you can also put some comment here. It will help someone in the future.

ann26:
Alright, working on it

django_project/base/models/project.py
ismailsunni: btw, @ann26 the travis also complains about it.

ann26: @ismailsunni is that because unidecode is not a default python package?

django_project/base/models/project.py
```diff
@@ -128,7 +129,8 @@ def save(self, *args, **kwargs):
         if not self.pk:
             words = self.name.split()
             filtered_words = [t for t in words if t.lower() not in STOP_WORDS]
-            new_list = unicode(' '.join(filtered_words))
+            #unidecode() takes Unicode data (e.g. special characters) and represents it in ASCII characters
```
ismailsunni: Space after # character for comment, and make sure the line is at most 80 chars

deployment/docker/REQUIREMENTS.txt
```diff
@@ -30,4 +30,6 @@ django-rosetta==0.7.0
 django-embed-video==1.0.0
 django-grappelli==2.7.3
 pytz
-django-allauth==0.25.2
\ No newline at end of file
+django-allauth==0.25.2
+#support special characters
```
ismailsunni: Space after # char

ann26:
Alright, I have fixed it

PR #405, opened 2017-03-29
Certification menu and certifying organisation page

### LINKED ISSUE #398: Create register certifying organisation view
#397 

create a register view to register a certifying organisation, including dropdown menu and details.

ann26:
fix #398 

includes certification drop down menu, certifying organisation registration page and list view.


meomancer:
hi @timlinux 
there is a blocker for anita. when we do migrate, there is error that occured.
Anita is using production database, but it seems base_project on database loses it's constraint?
i'm not sure, because i checked it and it doesn't has constraint.
```
HTTPError: HTTP Error 502: Bad Gateway
ERROR 2017-04-05 07:34:25,458 base 1 140397258737408 [u'ProgrammingError: there is no unique constraint matching given keys for referenced table "base_project"\n', u'
```

it seems base_project loses his constraint.
```
                                                       Table "public.base_project"
   Column    |          Type          |                         Modifiers                         | Storage  | Stats target | Description 
-------------+------------------------+-----------------------------------------------------------+----------+--------------+-------------
 id          | integer                | not null default nextval('base_project_id_seq'::regclass) | plain    |              | 
 name        | character varying(255) | not null                                                  | extended |              | 
 image_file  | character varying(100) | not null                                                  | extended |              | 
 approved    | boolean                | not null                                                  | plain    |              | 
 slug        | character varying(50)  | not null                                                  | extended |              | 
 private     | boolean                | not null                                                  | plain    |              | 
 description | character varying(500) |                                                           | extended |              | 
 owner_id    | integer                | not null                                                  | plain    |              | 
 gitter_room | character varying(255) |                                                           | extended |              | 
```

So we use fresh database now. Could you please check database on projecta? because i don't have access to projecta server.

PR #406, opened 2017-04-03
make pep8 compulsory

Gustry:


Gustry:
I'm just discovering that this plugin got flak8. Feel free to cose this PR if it's not relevant. I think we could improve flake8 maybe

.flake8
```diff
@@ -1,8 +1,8 @@
 [flake8]
 exclude = */docs/*,*/.tox/*,*/.venv/*
-max-line-length = 100
+max-line-length = 89
```
ismailsunni: Why is it set to 89? no 80?

.flake8
```diff
@@ -1,8 +1,8 @@
...
 
 # E12x continuation line indentation
 # E251 no spaces around keyword / parameter equals
 # E303 too many blank lines (3)
-ignore = E126,E251,E303
+ignore = E126,E251,E303,E125
```
ismailsunni: May be you can add the explanation of E125 like the others in the lines above

.flake8
Gustry: Ah yes, typo. I guess it should be 79, no ?

.flake8
Gustry: The same as line 5: `# E12x continuation line indentation`

ismailsunni: ah yes, I didn't see the `x` :D

.flake8
timlinux: Can you tweak this and then merge @Gustry ?

Gustry:
I'm moving from pep8 to flake8. I will merge this one soon @timlinux

cchristelis:
@Gustry, thanks for this.

Gustry:
We should add this flake8 test on every python repository. I will merge this PR today, Travis should be happy after that.

PR #435, opened 2017-06-12
return 404 for bad id certificate

### LINKED ISSUE #432: 500 on bad certificate number
# Problem

When I try and add a bad certificate number, then I get a 500, I should get a 404.
See:
http://staging.changelog.qgis.org/en/qgis/certificate/0246242/

# Proposed Solution

Return a 404

ann26:
fix #432 
peek 2017-06-12 17-31

### DIFF
--- a/django_project/certification/views/certificate.py
+++ b/django_project/certification/views/certificate.py
@@ -1,5 +1,5 @@
 # coding=utf-8
-from django.http import Http404
+from django.http import Http404, HttpResponseNotFound
 from django.views.generic import CreateView, DetailView
 from django.core.urlresolvers import reverse
 from braces.views import LoginRequiredMixin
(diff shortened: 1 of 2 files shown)


django_project/certification/views/certificate.py
```diff
@@ -125,7 +125,8 @@ def get_object(self, queryset=None):
         if queryset is None:
             queryset = self.get_queryset()
             certificateID = self.kwargs.get('id', None)
-            if certificateID:
+            certificate = Certificate.objects.all().values_list('certificateID', flat=True)
+            if certificateID in certificate:
```
cchristelis: why not rather use the `get or 404` pattern here? I think we should rather use the DB to search for a record than return all the objects and search in python...

timlinux:
Maybe rather than returning a 404 it would be better to show a big red 'X' and a note underneath saying that no certificate matching this number has been issued.

django_project/certification/views/certificate.py
ann26: ok, I've fixed it

ann26:
@timlinux ok, I fixed it
peek 2017-06-14 11-53

timlinux:
Perfect, thanks!

PR #446, opened 2017-06-14
use user real name instead of their username

### LINKED ISSUE #441: who is bob123?
# Problem

It is not easy to know who the user is by username, let's rather use the Human name.

# Proposed Solution 

We should rather display Bob Smith than bob123 when selecting users.

ann26:
fix #441 

peek 2017-06-14 12-37

### DIFF
--- a/django_project/certification/forms.py
+++ b/django_project/certification/forms.py
@@ -149,6 +149,7 @@ class CourseConvenerForm(forms.ModelForm):
         self.helper.layout = layout
         self.helper.html5_required = False
         super(CourseConvenerForm, self).__init__(*args, **kwargs)
+        self.fields['user'].label_from_instance = lambda obj: "%s <%s>" % (obj.get_full_name(), obj)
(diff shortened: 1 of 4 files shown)


django_project/certification/forms.py
```diff
@@ -149,7 +149,8 @@ def __init__(self, *args, **kwargs):
         self.helper.layout = layout
         self.helper.html5_required = False
         super(CourseConvenerForm, self).__init__(*args, **kwargs)
-        self.fields['user'].label_from_instance = lambda obj: "%s <%s>" % (obj.get_full_name(), obj)
```
Gustry: @ann26 You can set your python editor to allow only 79 characters, you will avoid these commits to fix travis ;-)

ann26: Ah yes! I didn't know that, thanks @Gustry :D

Gustry: screen shot 2017-06-14 at 10 18 04

sama sama

PR #455, opened 2017-06-23
change pin on map to point training centers

### LINKED ISSUE #440: Pins should be training locations not country
# Problem

We only see the pin for the country that the organisation is on. We actually want to see the training locations.

# Proposed Solution

Rather show pins for training locations.

ann26:
fix #440 

screenshot from 2017-06-23 12-24-33


timlinux:
LGTM - please resolve conflicts then merge

PR #462, opened 2017-07-06
change map with more detailed map

### LINKED ISSUE #460: Map at certain places sometimes does not load properly
# Problem
On training center details page, at certain cities, map sometimes does not load properly. I think it is because the data of the map itself.

ann26:
fix #460 

this map can shows details in more places than the one that was already implemented.
screenshot from 2017-07-06 13-19-49
screenshot from 2017-07-06 13-19-58


timlinux:
Good to apply this - its going to be a bit dodgy marketing wise to use ESRI map services

Please remove any other ESRI maps you might have used in other places too.

PR #467, opened 2017-07-10
fix map

ann26:
fix map issues when using chrome: sometimes map did not load.


timlinux:
Looks OK for me (without doing a deep review). Thanks.

Issue #394, opened 2017-03-15
Warning the domain name provided is not valid


timlinux:
I am getting lots of errors sent to me from the site like this:

```
Invalid HTTP_HOST header: u':8080'. The domain name provided is not valid according to RFC 1034/1035.

Request repr(): 
<WSGIRequest
path:/,
GET:<QueryDict: {}>,
POST:<QueryDict: {}>,
COOKIES:{},
META:{'CONTENT_LENGTH': '',
'CONTENT_TYPE': '',
'DOCUMENT_ROOT': '/etc/nginx/html',
'PATH_INFO': u'/',
'QUERY_STRING': '',
'REMOTE_ADDR': '200.75.240.139',
'REMOTE_PORT': '3669',
'REQUEST_METHOD': 'GET',
'REQUEST_URI': '/',
u'SCRIPT_NAME': u'',
'SERVER_NAME': '',
'SERVER_PORT': '8080',
'SERVER_PROTOCOL': 'HTTP/1.0',
'uwsgi.node': 'uwsgi',
'uwsgi.version': '2.0.14',
'wsgi.errors': <open file 'wsgi_errors', mode 'w' at 0x7f1228fab660>,
'wsgi.file_wrapper': <built-in function uwsgi_sendfile>,
'wsgi.input': <uwsgi._Input object at 0x7f120fe2a4c8>,
'wsgi.multiprocess': True,
'wsgi.multithread': False,
'wsgi.run_once': False,
'wsgi.url_scheme': 'http',
'wsgi.version': (1, 0)}>

```

We need to fix it so it stop spamming me :-(

PR #470, opened 2017-07-19
add admin

ann26:
Currently, I am not aware if the http error still persist ( #394 ).
So I think it would be good to add my email address in the admin, thus if the error happens again, I will be notified too.


timlinux:
Yeah I am still getting #394 errors all the mite - grateful if you can look into it and resolve it.

LGTM for your patch.

PR #475, opened 2017-07-25
[wip] modify certificate model and create transaction model for certificate

### LINKED ISSUE #468: Finalising certification
Basically we need a few more views to close the loop:

1) A view for an **organisation** owner which lists all certificates which have been issued but not paid for in their organisation (with a total after indicating the cost of outstaninding invoices)
Course type, Course date, User name & surname, Certificate link

http://staging.changelog.qgis.org/en/test/unpaidcertificates/list/

2) A view for a **project manager** which lists same as above but grouped by organisation

http://staging.changelog.qgis.org/en/test/unpaidcertificates/list/

```
QGIS Unpaid Certificates

Organisation: Foo

Course type, Course date, User name & surname, Certificate link
Course type, Course date, User name & surname, Certificate link
Course type, Course date, User name & surname, Certificate link
Course type, Course date, User name & surname, Certificate link
...
Course type, Course date, User name & surname, Certificate link

Total outstanding: USD 400

Org
```

3) A form for the **project manager** to register that he has recieved payment from an organisation owner. When he accepts the form, the received total is divided by the certificate unit price (to produce X) and then X items in the list of items unpaid for in the organisation are removed from the unpaid list (oldest first). A cross reference in the certificate record should be made to the payement record so you can see when the certificate was paid for

```
                        ___________                   
Organisation: |   Foo | v |
                        -----------

Amount outstanding: $400

Record payment: 

OK
```

4) a form that a project administrator can use to set the price for certificates
if the price is set to zero then the  view above  in organization is hidden (i.e. we hide the list of unpaid  certificates)

5) When recording payments for a batch of certificates, the unit prices should also be shown  so that if the project admin changes the price at the provject level, it is clear what the individual certificates costed

The revised schema (new entities in blue) might look like this:

screen shot 2017-07-12 at 4 01 13 pm

See @cchristelis  alternative approach below.

ann26:
fix #468 

in this PR
- [x] create certificate transaction model 
- [x] modify certificate model to include paid/unpaid status, cost, and transaction ID
- [x] create view to change the cost of the certificate for each course


ann26:
Hi @timlinux 

the current view is this:
screenshot from 2017-07-28 16-02-34
staff and project owner can see all organisations unpaid certificate, while the organisation owner can only see their organisation unpaid certificate,

In the model, there's variable `cost` now, which will be use to calculate how much cost is still unpaid for each organisation. However, I am not sure where to set this? I mean, where would the organisation owner/project owner supply this value? Should it be the same for all organisation within the project?

ann26:
@timlinux here's to change the cost of the certificate:
the cost of the certificate is the same if they are in the same course
peek 2017-08-07 16-17

timlinux:
Thats looking good @ann26 
A couple of comments:

1) Maybe it more logical to have the unpaid certificates shown inside the org landing page instead of on the org list page.
2) For the certificate payment, @cchristelis 's idea (which I like) is to let the org owner use a 'top up' system. Each time a certificate is issued, we reduce their credit by the cost of one certificate. We sell the user popups in multiples of one certificate cost so e.g. 'buy 100 credits' - user pays 100 x certificate cost for the transaction.
3) The certificate cost should be set / a property of the project and only changeable by the project admins. Org admins will not be allowed to set their own cost rate.
4) We should show the current balance to the org owner each time a new certificate is issued and not allow a cert to be issued if there is no accompanying transaction.
5) We need to set up PayPal for the payments. There should be a PayPal API key set on a per-project basis so that popups for project x go to the account of project x.
6) For the button color accents, same advice goes as for your other PR's

ann26:
Hi @timlinux ,
If we use credit system, will the unpaid certificates view be unnecessary? Since when issuing the certificate, it will also deduct the credit? Maybe we can have a remaining credit page?

PR #478, opened 2017-07-26
fix edit organisation and pending navigation

### LINKED ISSUE #477: Editing a certifying organization removes it from view and does not show it in moderation
licecap

ann26:
fix #477 

This PR fixes:
- [x] Edit organisation, so it won't disappear from list view
- [x] Pending navigation, so it will appear in all details page in certification app

peek 2017-07-26 10-56

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
     - DJANGO_SETTINGS_MODULE=core.settings.test_travis
 
 sudo: false
+dist: precise
 
 python:
   - '2.7'
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ env:
 
 sudo: false
 
+dist: precise
+
 python:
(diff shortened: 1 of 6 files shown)


ann26:
Hi @timlinux in this PR, the travis checks were failed for some commits, and I found out that is due to travis updated their environment to use `dist: trusty` as default (previously they used `dist: precise` as default). In the last commit, I add `dist: precise` to the travis.yml , and travis check passed.

timlinux:
Does your change here mean that organization changes won't be pending approval? I think it is ok if the organization owner does edits that need no approval but non staff and organization owners should not be able to change the organization.

ann26:
@timlinux yes, so only when the organisation is created, needs approval from staff. And yes, the only people who can edit the organisation are staff and organisation owners.

timlinux:
Ok good thanks!

PR #482, opened 2017-07-31
add course date validation

ann26:
This PR add validation for input course start date and end date
- [x] When course start date is available, the input course end date can only be after course start date
- [x] When course end date is available, the input course start date can only be before course end date
- [x] Datepicker automatically close when a date is selected

peek 2017-07-31 10-31


timlinux:
Good fix thanks @ann26

PR #489, opened 2017-08-02
use real name for organisation owners

ann26:
In the organisation add/edit page:

for organisation owners:
screenshot from 2017-08-02 16-20-06


timlinux:
Thanks @ann26 !

PR #492, opened 2017-08-07
fix navigation bar on the certificate validation page

ann26:
Fix navigation bar in the validate certificate page

screenshot from 2017-08-07 18-12-27


timlinux:
Lovely thanks.

PR #493, opened 2017-08-08
add user automatically as default organisation owner

ann26:
When a user create an organisation, he will be automatically chosen as an organisation owner.

screenshot from 2017-08-08 16-12-37


timlinux:
Great improvement thanks

PR #515, opened 2017-08-29
certificate credit systems

### LINKED ISSUE #468: Finalising certification

ann26:
fix #468 

This PR:
- [x] modify project model to include certificate cost
- [x] modify organisation model to include certificate credits
- [x] every time a certificate is issued, the certificate credits of the organisation will be deducted by 1

It still allows certificate credits to be minus.


django_project/certification/views/certificate.py
```diff
@@ -97,6 +97,15 @@ def form_valid(self, form):
 
         try:
             super(CertificateCreateView, self).form_valid(form)
+
+            # Update organisation credits every time a certificate is issued.
+            organisation = \
+                CertifyingOrganisation.objects.get(
+                    slug=self.organisation_slug)
+            remaining_credits = organisation.credits_certificate - 1
```
cchristelis: This one cannot be hard-coded. Let's put it in the project settings model.

ann26:
@timlinux and @cchristelis 
here is update:
In here, the example is a certificate cost 5 credit (editable in project model).
When the credits is insufficient, icons changed to `top up icons`
peek 2017-09-08 18-33

when credits is topped up, user may choose which certificate they want to pay
peek 2017-09-08 18-50

`top up` icons and `pay with credits` icons is not working yet

ann26:
Pay unpaid certificates using credits that has just been top up
peek 2017-09-11 12-12

ann26:
current top up view:
screenshot from 2017-09-11 13-13-41

timlinux:
@ann26 I'm going to merge this as we have another user who wants to test the system. Please come back after and add tests ok?

ann26:
ok @timlinux

PR #519, opened 2017-09-07
fix certificate numbering

ann26:
fix certificate numbering so that it'll count the certificate for every project:
screenshot from 2017-09-07 12-19-10


cchristelis:
LGTM.

PR #534, opened 2017-09-19
fix cache certificate

### LINKED ISSUE #533: Print certificate using cache does not work
# Problem
From a user: he tried to print the certificate, but it does not work.
Here's his screenshot:
sc1
sc2

# Solution
fix it

ann26:
fix #533 
peek 2017-09-19 16-18


timlinux:
Thanks @ann26

PR #542, opened 2017-09-22
add sponsorship amount paid

### LINKED ISSUE #220: Sponsorship period should also include the amount sponsored for that period
It could be that the sponsor gives more than the minimum amount for that sponsorship level, so we should also record the exact amount given.

ann26:
fix #220

form:
screenshot from 2017-09-22 15-16-55
screenshot from 2017-09-22 15-17-08

sponsor list:
the amount sponsored is displayed when it is available.
screenshot from 2017-09-22 15-15-52

sponsorship period details:
when the amount sponsored is available:
screenshot from 2017-09-22 15-16-04

when the amount sponsored is not available:
screenshot from 2017-09-22 15-16-14

sponsor details: amount sponsored is displayed when it is available.
screenshot from 2017-09-22 15-45-54

### DIFF
--- a/django_project/changes/forms.py
+++ b/django_project/changes/forms.py
@@ -254,7 +254,7 @@ class SponsorshipPeriodForm(forms.ModelForm):
             'sponsorship_level',
             'start_date',
             'end_date',
-            'amount_paid',
+            'amount_sponsored',
             'currency'
         )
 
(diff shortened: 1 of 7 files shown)


timlinux:
@andreasneumann could you review this?

timlinux:
@ann26 I think we must just go ahead and merge this - we can tweak again if @andreasneumann has any further comments

andreasneumann:
Looks ok to me. Thanks a lot!

Would this also be used in the automatically generated invoice?

andreasneumann:
Separate currency also makes sense, because in some cases sponsor want to pay in their local currency.

However, € should be the default currency. >90% of our sponsors us this currency.

ann26:
Hi @andreasneumann I added the default currency value (EUR) to the model

timlinux:
@andreasneumann wrote:

> Would this also be used in the automatically generated invoice?

I think we can do that - @ann26 can you work with @jeanabdesmidt to get https://github.com/kartoza/projecta/pull/272 merged then we should be able to do the full workflow for manual sponsorship management. When stripe support comes, we can automate the whole process...

ann26:
@timlinux will do

PR #549, opened 2017-09-25
add send email to all attendees with paid certificate

ann26:
fix #545 

icon in course details page:
screenshot from 2017-09-25 17-29-37

confirm page:
screenshot from 2017-09-25 17-24-27

gif:
peek 2017-09-25 17-26

email content:
```
Dear Jane Austen,

Congratulations!
Your certificate from the following course has been issued.

Course type: Coding for Dummies
Course date: 11 July 2017 to 19 December 2017
Training center: training center B
Certifying organisation: OrganisationX

You may print the certificate by visiting:
http://0.0.0.0:61202/en/x-project/certifyingorganisation/xproject-organisationx/course/xproject_coding-for-dummies_2017-07-11-2017-12-19/print/37/

Sincerely,
Anita Hapsari
```

PR #554, opened 2017-09-27
tidy up string format

ann26:


timlinux:
Thanks for doing this - see https://github.com/kartoza/projecta/pull/549#issuecomment-332738761

ann26:
done @timlinux

PR #569, opened 2017-10-06
show amount sponsored and retrofill data

### LINKED ISSUE #568: show actual amount sponsored next to each organisation name
In the sponsorship periods view, it would be good to show actual amount sponsored next to each organisation name

ann26:
fix #568 

this PR:
- [x] shows amount sponsored in sponsorship period list
- [x] retrofill amount sponsored data with sponsorship level value for existing objects

screenshot from 2017-10-06 11-01-03


timlinux:
Perfect thanks. @andreasneumann we are going to pre-populate the amounts with the sponsorship level. If you can update them to the exact amounts when the amount was different then we can make reports which show the exact sponsorship amounts.

LGTM @ann26

PR #588, opened 2017-10-11
include all test in project in travis

### LINKED ISSUE #559: Have travis run test for the whole project
# Problem
Since travis updated its environment, some of the old tests (that was passed) are failing. So they are excluded. 

# Solution
Maybe we should investigate to include all test in travis.

ann26:
fix #559


timlinux:
Oh sorry I also did this on my side. Also enabled collect static when travis runs.

ann26:
ok @timlinux no problem

PR #590, opened 2017-10-11
Make travis run all tests

### LINKED ISSUE #559: Have travis run test for the whole project

timlinux:
This PR re-enables all tests for travis. See #559


timlinux:
OK all tests running on travis and passing now.

screen shot 2017-10-11 at 20 33 22


CC @ann26

PR #594, opened 2017-10-13
fix display icon

ann26:
before fix (in safari and chrome):
screen-shot-2017-10-13-at-09 24 36

after fix (tested in chrome):
screenshot from 2017-10-13 14-38-57


timlinux:
LGTM thanks

PR #614, opened 2017-10-20
add certification managers

### LINKED ISSUE #596: We need to add a role for project managers
For any project the project owner should be able to define zero or more project managers. Managers would be like staff role in the built in django permissions system except their rights would only be applied on the selected project. They should be able to approve pending entries / organisations etc.

Please add something like this to the project model:

screen shot 2017-10-13 at 11 12 49

ann26:
fix #596 

- [x] This PR added certification manager role in the project model. 
- [x] The certification manager will receive a notification email when there is a new organisation registered (will be addressed in different PR)
- [x] The certification manager has the same permissions with project owner in certification, such as:
 * Create/delete/update certifying organisation, training center, course, course attendee, course type, and course convener, attendee
 * Approve certifying organisation
 * Issue certificates
 * Regenerate certificates
 * Send emails to participant
- [x] Certification manager is **not** able to update/delete project

Form:
screenshot from 2017-10-20 10-59-43
screenshot from 2017-10-20 11-00-54
screenshot from 2017-10-20 11-01-12
screenshot from 2017-10-20 11-02-08
screenshot from 2017-10-20 11-03-18
screenshot from 2017-10-20 11-03-45


timlinux:
Thanks @ann26 LGTM

CC @geomenke

PR #616, opened 2017-10-23
add sponsor feed

### LINKED ISSUE #612: We need a feed of sponsors into the QGIS.org site
Consider having an RSS feed or a url that will return zip containing the content

TODO Tim: Website: Include sponsors list from http://changelog.qgis.org/en/qgis/sponsors/list/ into http://www.qgis.org/en/site/about/sponsorship.html so Andreas doesn’t have to do twice the work (make an entry in the changelog application and manually update the website in Github)

ann26:
fix #612 

On firefox:
screenshot-2017-10-23 rss sponsor of qgis project

### DIFF
--- a/django_project/changes/feeds/sponsor.py
+++ b/django_project/changes/feeds/sponsor.py
@@ -109,7 +109,7 @@ class RssSponsorFeed(Feed):
             'currency': item.currency,
             'amount_sponsored': item.amount_sponsored,
         }
-        
+
         descriptions = \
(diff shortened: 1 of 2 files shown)


timlinux:
@rduivenvoorde and @andreasneumann is that going to work for you? I will merge this and deploy it for you to play with it and give feedback.

timlinux:
LGTM thanks @ann26

PR #625, opened 2017-10-27
make sponsor programme field longer

### LINKED ISSUE #624: Make the field length for sponsorship programme longer
Trying to paste in long content into the QGIS sponsorship programme fails - please fix by making the field length larger

ann26:
fix #624 

I copied the content from here as example: http://www.qgis.org/en/site/getinvolved/governance/sponsorship/sponsorship.html

screencapture-0-0-0-0-61202-en-qgis-sponsorship-programme-1509077004929


timlinux:
Thanks - looks good!

PR #635, opened 2017-11-29
add custom domain

### LINKED ISSUE #570: We need to support user domains
When a user domain is set, only projects owned by that user should show on the home page. We should provide the user with instructions on using a CNAME DNS record to point to our site. We need to take care of allowed_hosts and potential security implications (see https://stackoverflow.com/a/30551895).

The linking of projects to custom domains should be supported in two ways:

1) **Organisation level** (we need to introduce an organisation model for this). In this model the projecta landing page (e.g. by clicking on the home icon) would should the gallery of all projects registered by that organisation. e.g. http://projects.fooorg.com would show only Foo Organisations projects.

2) **Project level**: when the cname entry points to a project the project page should be the landing page, the home page icon should be removed in the menu bar making the project the top level navigation point.

This should be a paid for service (on a subscription model) so we need a sign up form which provides the needed instructions and the middleware accepting cname hosts should also make a check to see if the user owning that domain is subscribed. If not a message inviting the user to subscribe should be show (if they are the org / project owner) or otherwise a generic message should be show indicating that the URL is invalid.

Here is a workflow overview:

customdomainhandling

### LINKED ISSUE #603: We need an organisation model
See #570

The organisation model should let projects be associated to an organisation. It should have its own owner and manger lists

ann26:
fix #570 and fix #603 

# [ WIP ]

This PR:
- created custom domain model where subscribed user and user domain is saved
- when the user and domain are approved, it will write the user domain to `server-name.txt` which is included in nginx so the domain will be included in the allowed hosts.

@timlinux and @cchristelis is this the right approach?

in the admin panel:
screenshot from 2017-11-29 18-03-07
screenshot from 2017-11-29 18-03-20


django_project/base/models/custom_domain.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+
+
+ROLE = (
+    ('project', 'project'),
+    ('organisation', 'organisation'),
+)
+
+
+class CustomDomain(models.Model):
```
timlinux: I think we should just call it 'Domain' and then create a Domain entry for e.g. http://projecta.kartoza.com which is the default domain that all new projects get added to that do not have an explicitly set domain. Then add a foreign key reference to Domain from both the organisation and project models.

deployment/sites-enabled/default.conf
```diff
@@ -20,6 +20,7 @@ server {
     # the domain name it will serve for
     server_name staging.projecta.kartoza.com changelog.linfiniti.com changelog.kartoza.com changelog.qgis.org changelog.inasafe.org staging.changelog.qgis.org;
     charset     utf-8;
+    include /home/web/media/server-name.txt
```
timlinux: I would imaging that since we will on a dedicated server using * for server name might be a simpler approach and then let projecta handle whether the hostname in the header of the request if valid or not.

timlinux: (that we we do not need to use server-name.txt tricks)

django_project/base/models/custom_domain.py
```diff
@@ -0,0 +1,57 @@
...
+        blank=True
+    )
+    approved = models.BooleanField(
+        help_text=_('Whether this user domain has been approved for use yet.'),
+        default=False
+    )
+
+    class Meta:
+        ordering = ['user']
+        unique_together = ['user', 'custom_domain']
```
timlinux: I think the domain would need to be independently unique - otherwise what will happen if two users try to register the same domain?

django_project/base/models/custom_domain.py
```diff
@@ -0,0 +1,57 @@
...
+        null=True,
+        blank=True
+    )
+    approved = models.BooleanField(
+        help_text=_('Whether this user domain has been approved for use yet.'),
+        default=False
+    )
+
+    class Meta:
+        ordering = ['user']
```
timlinux: lets order by the domain name

timlinux:
@ann26 Referring back to my original ticket at #635 I think it makes more sense to make an atomic PR that adds an organisation model and associates each project with each organisation. By default we can associate projects with the 'global' organisation until the are assigned to a specific organisation.

Then each organisation and project will get associated with a domain.

timlinux:
@ann26 can you also take a look at this PR from @meomancer and review whether the code will fit within the framework we are setting up here:

https://github.com/kartoza/projecta/pull/361

ann26:
ok @timlinux will do

ann26:
Hi @timlinux , looking at @meomancer 's PR (#361) I think the code can fit our framework with a bit of modification. 

I tried it in my local and using it we can assign additional administrators that can have permissions the same as the owner in the project, as shown below:
screenshot from 2017-12-21 20-35-16
So in above example case, I am (not a staff) the owner of Inasafe project and an administrator of XProject project. I have the same permissions as the owner of XProject eventhough I am not the owner of it. 

The app also includes custom template tag that we can implement in template, e.g. `the_project|is_project_administrator:user` 

Maybe the Project Collaborators can be the Managers of each app (e.g. changelog, sponsorship, certification) ?

In conclusion, I think using permission app, we can manage permissions separately from the project model and its custom template tag can simplify the template code.

ann26:
Hi @timlinux , here is update:
- [x] Added organisation model
- [x] Added organisation create and list view
- [x] The existing projects are assigned to default organisation (Kartoza)
- [x] Added domain model
- [x] Added domain create and list view
- [x] Domain list view only available for staff
- [x] Added domain check middleware
- [x] Changed nginx conf server name to `*` (wildcard)
- [x] When a domain is registered with role as project, the site will point to project page
- [x] When a domain is registered with role as organisation, the site will point to home page (project list page) with only showing the project in that organisation.
- [x] Added domain check middleware test
- [x] Update existing tests

Gifs:
- Organisation create view
peek 2018-01-03 17-07

- Organisation list and pending view
peek 2018-01-03 17-08

- Domain create view
User may click on `register a domain` from navigation.
peek 2018-01-03 17-32

Or may click on `set up custom domain` button on project page
peek 2018-01-03 17-42

- When domain is not registered or not approved yet
peek 2018-01-03 17-35

- Domain list view (only available to staff users)
peek 2018-01-03 17-36

- Domain as organisation role
peek 2018-01-03 17-38

- Domain as project role
peek 2018-01-03 17-40


Travis is still failing, I am still investigating it.

timlinux:
Wow @ann26 beautiful work! Please merge when you have travis passing and then deploy it to staging so I can play with it a bit.

PR #636, opened 2017-12-01
Hide organisational private related details from non-logged in user.

### LINKED ISSUE #629: Organisation private details should be visible only to organisation owner / manager
Here is a simple example of what a non logged in user can see

screen shot 2017-11-14 at 11 32 34

This kind of information should be only shown to organisation owners or managers.

sonlinux:
Fix #629

Do not show organisational related private details to non organisational owners or staff or admin users.

screenshot from 2017-12-01 15 14 55

### DIFF
--- a/django_project/certification/templates/course/detail.html
+++ b/django_project/certification/templates/course/detail.html
@@ -71,47 +71,48 @@
     </div>
 
     <div class="row" style="margin-left: inherit; margin-top: 15px;">
-    <div class="details col-lg-10" style="width: 60%">
-        <span class="grey-italic col-lg-3">Course name</span><span class="col-lg-9">{{ course.course_type.name }} ({{ course.start_date }} to {{ course.end_date }})<br/></span>
-        <span class="grey-italic col-lg-3">Course type</span><span class="col-lg-9">{{ course.course_type }}<br/></span>
-        <span class="grey-italic col-lg-3">Course convener</span>
-        {% if course.course_convener.user.first_name %}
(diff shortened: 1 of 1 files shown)


timlinux:
Hi @sonlinux 

Thanks for this patch. I was really intending in #629 that you step back and review the whole workflow rather than just the single example I gave and make a determination on each view as to whether that view should be visible to all users, certifying organisation managers or no only admins. A good way to do this would be to produce a table of all certification views and a column indicating who should be able to see that view. Do the table in the original ticket and we can review it. When agreement is reached, you can implement the rules as you have done in this PR.

Could I also suggest that you implement a custom template tag to use when evaluating whether a block should be visible rather than having and extended set of if/else clauses in the template?

sonlinux:
Hi @timlinux 
I'll do just that. 
Thank you.

sonlinux:
Hi @timlinux I have worked on the view table we can review when you can, I have shared it on google drive. Thanks.

sonlinux:
Hi @timlinux 
I have reviewed the certification views and they are abiding to the permissions we reviewed in the table. 
So I opened a seperate ticket for the template tag so that I work on it seperately.

Could I also suggest that we open a ticket for properly catching the error on validating an empty certificate entry ill work on it.
As at now when I search for a null certificate ID I get thrown a page not found excpetion.

screenshot from 2017-12-05 10 28 40

screenshot from 2017-12-05 10 28 51

Thank you @timlinux

timlinux:
LGTM thanks @sonlinux

PR #652, opened 2017-12-20
add user permission template tag.

### LINKED ISSUE #640: Implement a custom template tag.
Implement a custom template tag to use when evaluating whether a block should be visible rather than having an extended set of if/else clauses in the template.

sonlinux:
Fix #640 
Add custom user permission templatetag.


django_project/base/templates/project/new_detail.html
```diff
@@ -1,6 +1,7 @@
 {% extends "project_base.html" %}
 {% load staticfiles %}
 {% load custom_markup %}
+{% load  user_perm %}
```
ann26: I think it is better to write user_permissions instead of user_perm, to make it more clear

django_project/base/templatetags/user_perm.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter(name='has_perm', is_safe=True)
```
ann26: this too, it is better to write `has_permissions`

django_project/base/templatetags/user_perm.py
```diff
@@ -0,0 +1,12 @@
...
+def has_perm(user, perm_name):
```
ann26: this too, permissions_name

django_project/base/templatetags/user_perm.py
```diff
@@ -0,0 +1,12 @@
...
+    perm = user.user_permissions.get(codename=perm_name)
```
ann26: this too, it is better to user `permissions`

django_project/base/templatetags/user_perm.py
```diff
@@ -0,0 +1,12 @@
...
+    if perm in user.user_permissions.all():
```
ann26: also this, `permissions`

timlinux:
@sonlinux There is a problem in this approach - it does not cater for the fact that a user may have permissions (e.g. changelog manager) in one project but not another. We cant use django build in permissions framework in this case but need to filter the permissions by the project rather.

timlinux:
I'm going to close this @sonlinux - there are other ways we can do this using a view method or the view context I think....

PR #656, opened 2017-12-22
include 'http' to permalink

### LINKED ISSUE #627: Permalink is missing the HTTP protocol
screen shot 2017-10-30 at 08 41 15

The HTTP protocol is missing from the permalink. So when we copy/paste, it creates a link not clickable like this:
changelog.qgis.org/en/qgis/version/3.0.0/#remove-trusted-status-from-plugin-manager

sonlinux:
Fix  #627 

screenshot from 2017-12-22 14-06-24

### DIFF
--- a/django_project/changes/templates/entry/includes/entry_detail.html
+++ b/django_project/changes/templates/entry/includes/entry_detail.html
@@ -31,7 +31,7 @@
 
             <a class="btn btn-default btn-mini tooltip-toggle"
                data-title="Get Permalink"
-               onclick="copyToClipboard('{{ request.get_host }}{{ request.path }}#{{ entry.slug }}')">
+               onclick="copyToClipboard('http://' + '{{ request.get_host }}{{ request.path }}#{{ entry.slug }}')">
                 <span class="glyphicon glyphicon-link"></span>
             </a>
         </div>
--- a/django_project/changes/templates/entry/includes/entry_detail.html
+++ b/django_project/changes/templates/entry/includes/entry_detail.html
@@ -31,7 +31,7 @@
 
             <a class="btn btn-default btn-mini tooltip-toggle"
                data-title="Get Permalink"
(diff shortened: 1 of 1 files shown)


Gustry:
@sonlinux This file is plain HTML, I don't think you need concatenation in that case.

sonlinux:
>@sonlinux This file is plain HTML, I don't think you need the '+' concatenation operator in that case. I think it should be exactly like the # on the same line.

Updated @Gustry

Gustry:
Sorry, I have another comment.
In the `request` documentation of django, there is `scheme`, it will give you the correct protocol used for the request (ftp, http, ...) . It will even switch between `https` or `http` if needed/available for an HTTP request.
So you should use the `request.scheme` property. Try it and tell me if it works.

Then for Travis, I will let @ann26 check for that

Gustry:
There is also `get_raw_uri` will should do all of them in one call before the slug

ann26:
hi @Gustry and @sonlinux , maybe travis updated its environment? I ran travis in the branch that previously passed the test and now it is failed :(

sonlinux:
>Sorry, I have another comment.
In the request documentation of django, there is scheme, it will give you the correct protocol used for the request (ftp, http, ...) . It will even switch between https or http if needed/available for an HTTP request.
So you should use the request.scheme property. Try it and tell me if it works.

Hi @Gustry, 
the ```request.scheme``` property works fine.

sonlinux:
Hi @Gustry 
Please take a another review on this PR when you can.
Thanks

sonlinux:
thanks @Gustry 
But we need to fix travis environment before merge.

Gustry:
The problem is already in develop: https://github.com/kartoza/projecta/commits/develop so it's fine IMHO

sonlinux:
Alright, ill go ahead and merge,

PR #665, opened 2018-01-09
Fix sidepanel display

### LINKED ISSUE #664: Side panels are rendering really ugly
screen shot 2018-01-09 at 08 51 25

@sonlinux can you please investigate and fix - ping me if it is not obvious from the screenie what the issue is

sonlinux:
Fix #664 

action-display

Hi @timlinux 
How is the looks for the sidepanel content on the screenie above?
Thanks


timlinux:
That looks much better thanks! LGTM

Issue #657, opened 2018-01-03
We need to implement the lessons app


timlinux:
It should implement the following models / schema

![lessonsapp](https://user-images.githubusercontent.com/178003/34536119-e0167c36-f0cc-11e7-9a06-a7ca9907faa3.png)

timlinux:
CC for your interest @charlotte-morgan

Charlotte-Morgan:
@timlinux - thanks for tagging me here. I hope CRUD docs are not the same as CRAP maps :)

timlinux:
@Charlotte-Morgan :-) 

CRUD - Create, Read, Update, Delete

ismailsunni:
More task:
- [ ] new project role here for lessons_manager...and make it editable by staff, lessons manager or project owner...

ismailsunni:
Tasks: https://docs.google.com/spreadsheets/u/1/d/1hpCL1ZreIDWBreaF_nWuqwrQROV4bRmAS3tcvaCefyk/edit?usp=drive_web

ismailsunni:
Hi @timlinux 
>  new project role here for lessons_manager...and make it editable by staff, lessons manager or project owner...

Can we just use group permission for this? And check the permission for CRUD-ing lesson?

timlinux:
> Can we just use group permission for this? And check the permission for CRUD-ing lesson?

I think it is better model it as I described @ismailsunni  - the permissions are specific to a particular project or organisation - IIRC the groups in django will be global across all projects which wont work.

@Gustry @ismailsunni @myarjunar Sorry I didnt make it clear but we should copy the interactions from changelogs: Show the section / lesson list in the side box, clicking on the box will take you to a menu of all sections / lessons, clicking on a specific lesson will take you to a read only render of that lesson

![lessons](https://user-images.githubusercontent.com/178003/35099360-63e32958-fc60-11e7-99a6-a4e23b0cb49f.gif)


At the top right corner of the read only lesson render (which should look just like the worksheet example above) should show three icons:

1. print icon (print to pdf using js pdf generator)
2. edit pencil icon (available to lesson managers, project admins and staff only) - takes you to the lesson form to tweak it
3. remove icon (available to lesson managers, project admins and staff only) - lets you remove the lesson

timlinux:
Here are the mappings of the worksheet fields to the content on the rendered worksheet:

[...]

PR #666, opened 2018-01-10
[WIP] Add lesson app model.

ismailsunni:
Issue: #657 
cc @timlinux @Gustry @myarjunar


django_project/lesson/__init__.py
```diff
@@ -0,0 +1,2 @@
+# coding=utf-8
+"""Docstring here."""
```
timlinux: Lesson app for projecta

django_project/lesson/admin.py
```diff
@@ -0,0 +1,40 @@
...
+from django.contrib import admin
+from lesson.models.answer import Answer
+from lesson.models.further_reading import FurtherReading
+from lesson.models.section import Section
+from lesson.models.specification import Specification
+from lesson.models.worksheet import Worksheet
+from lesson.models.worksheet_question import WorksheetQuestion
+
+
+class AnswerAdmin(admin.ModelAdmin):
```
timlinux: shouldn't these get at least a 

pass

in their bodies?

django_project/lesson/models/answer.py
```diff
@@ -0,0 +1,45 @@
...
+import logging
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from lesson.models.worksheet_question import WorksheetQuestion, Worksheet
+logger = logging.getLogger(__name__)
+
+
+class Answer(models.Model):
+    """Answer lesson model."""
```
timlinux: Each lesson can have zero or more questions associated with it. The question will have one or more answers and the answer will have an indcation if it is a correct answer or not. One question could have more than one correct answers (or no correct answers).

django_project/lesson/models/further_reading.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from lesson.models.worksheet import Worksheet
+
+logger = logging.getLogger(__name__)
+
+
+class FurtherReading(models.Model):
+    """Further reading lesson model."""
```
timlinux: This is a forward link for the topic so that the lesson reader can find more info about the topic.

django_project/lesson/models/further_reading.py
```diff
@@ -0,0 +1,37 @@
...
+
+    worksheet = models.ForeignKey(Worksheet)
+
+    text = models.CharField(
+        help_text=_('Text of the further reading.'),
```
timlinux: Further reading text

django_project/lesson/models/further_reading.py
```diff
@@ -0,0 +1,37 @@
...
+        max_length=200,
+    )
+
+    link = models.CharField(
+        help_text=_('Link of the further reading.'),
```
timlinux: Further reading link

django_project/lesson/models/further_reading.py
timlinux: This should be null false blank false.

django_project/lesson/models/further_reading.py
timlinux: This can be null true blank true - better to explicitly declare every field as to whether it can be null and blank.....

django_project/lesson/models/further_reading.py
```diff
@@ -0,0 +1,37 @@
...
+        max_length=200,
+    )
+
+    # noinspection PyClassicStyleClass.
+    class Meta:
+        """Meta class for Course attendee."""
```
timlinux: cut and paste error?

django_project/lesson/models/section.py
```diff
@@ -0,0 +1,55 @@
...
+from django.utils.translation import ugettext_lazy as _
+from unidecode import unidecode
+from core.settings.contrib import STOP_WORDS
+from django.utils.text import slugify
+import logging
+logger = logging.getLogger(__name__)
+
+
+class Section(models.Model):
+    """Section lesson model."""
```
timlinux: A section is a grouping for one or more lessons on a common theme e.g. digitising. Sections can be ordered so that the learner goes through the sections in the correct sequence.

django_project/lesson/models/section.py
```diff
@@ -0,0 +1,55 @@
...
+
+    section_number = models.IntegerField(
+        help_text=_('Section number.'),
```
timlinux: please set null and blank properties for all fields...

django_project/lesson/models/section.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    name = models.CharField(
+        help_text=_('Name of lesson.'),
+        max_length=200,
+    )
+
+    notes = models.TextField(
+        help_text=_('Name of lesson.'),
```
timlinux: copy and paste error

django_project/lesson/models/section.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Section(models.Model):
+    """Section lesson model."""
+
+    section_number = models.IntegerField(
+        help_text=_('Section number.'),
+    )
+
+    name = models.CharField(
+        help_text=_('Name of lesson.'),
```
timlinux: Name of section

django_project/lesson/models/section.py
timlinux: Section notes.

django_project/lesson/models/specification.py
```diff
@@ -0,0 +1,48 @@
...
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from lesson.models.worksheet import Worksheet
+
+logger = logging.getLogger(__name__)
+
+
+class Specification(models.Model):
+    """Worksheet lesson model."""
```
timlinux: A specification is a particular requirement for a task that the learner is required to do in the lesson.

django_project/lesson/models/specification.py
```diff
@@ -0,0 +1,48 @@
...
+
+    worksheet = models.ForeignKey(Worksheet)
+
+    specification_number = models.IntegerField(
+        help_text=_('Specification number.'),
```
timlinux: Used to order the specifications for a lesson into the correct sequence

django_project/lesson/models/worksheet.py
```diff
@@ -0,0 +1,96 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Worksheet(models.Model):
+    """Worksheet lesson model."""
+
+    section = models.ForeignKey(Section)
+
+    module = models.CharField(
+        help_text=_('Name of module.'),
```
timlinux: Name of worksheet

django_project/lesson/models/further_reading.py
ismailsunni: Both are `False` by default according to this https://docs.djangoproject.com/en/2.0/ref/models/fields/#null

Should I make it explicit? Actually I was writing `null=False, blank=False` everywhere, and delete them since it's already the default value.

ismailsunni: Ah I just read your next comment, I will make it explicit.

django_project/lesson/models/further_reading.py
ismailsunni: Yes :(

django_project/lesson/models/further_reading.py
ismailsunni: Btw, few days ago I read that it's better not use `null=True` for CharField or TextField (unless we need it to be unique). https://docs.djangoproject.com/en/1.8/ref/models/fields/#null

django_project/base/templates/project/new_detail.html
```diff
@@ -194,6 +194,36 @@ <h6><span>No teams are defined, but you can </span>
...
+                                <span>
+                                    <a href="">
+                                        {{ section.name }}
+                                    </a>
+                                </span>
+                            </h5>
+                        {% endfor %}
+                    </div>
+                    {% else %}
+                    {% if user.is_staff or user == project.owner %}
```
timlinux: We should add a new project role here for lessons_manager...and make it editable by staff, lessons manager or project owner...

django_project/lesson/forms/section.py
```diff
@@ -0,0 +1,54 @@
...
+            'notes'
+        )
+
+    def __init__(self, *args, **kwargs):
+        self.project = kwargs.pop('project')
+        self.helper = FormHelper()
+        layout = Layout(
+            Fieldset(
+                _('Section details'),
+                Field('section_number', css_class='form_control'),
```
timlinux: I dont think it makes sense to let the user set a section number - it should be calculated dynamically. The section order is intended to be mostyly invisible to the user and will determine the sort order for the sections (and numbers need not be sequential).

timlinux: Have a look at how @dimasciput did the sortable implementation for changelog categories for an example of what we are looking for here...

django_project/lesson/views/__init__.py
```diff
@@ -0,0 +1,2 @@
+# coding=utf-8
+"""Docstring here."""
```
timlinux: Need a better docstring....

django_project/lesson/forms/section.py
ismailsunni: I just checked the form for creating new category, it also has the sorting number
screen shot 2018-01-11 at 15 20 23

But, I understand by ordering (after seeing the ordering in http://changelog.qgis.org/en/inasafe/category/order/ )

Can we do it later? So that we can divide the work?

django_project/lesson/views/__init__.py
ismailsunni: Done.

django_project/base/templates/project/new_detail.html
ismailsunni: I will add to the ticket, see: https://github.com/kartoza/projecta/issues/657#issuecomment-356860151

django_project/lesson/admin.py
ismailsunni: Added.

django_project/lesson/models/further_reading.py
timlinux: Thanks

django_project/lesson/forms/section.py
timlinux: yes fine to do later thanks

PR #686, opened 2018-01-17
Section View

ismailsunni:
- CRUD for section
lesson_app

- Ordering for section (Sorry @myarjunar , it should be yours, I look at the wrong line in the excel)
section_ordering

- Unit test

cc @Gustry @myarjunar I have a migration in this PR.


django_project/lesson/views/section.py
```diff
@@ -126,7 +132,329 @@ def get_queryset(self):
...
+        return context
+
+    def get_queryset(self):
+        """Get the queryset for this view.
+
+        :returns: Queryset for All Sections
+        :rtype: QuerySet
+        """
+
+        qs = Section.objects.all()
```
Gustry: This is already the default, no? Is the view not based on the model already?

django_project/lesson/views/section.py
```diff
@@ -126,7 +132,329 @@ def get_queryset(self):
...
+        sections_json = request.body
+
+        try:
+            section_request = json.loads(sections_json)
+        except ValueError:
+            raise Http404(
+                'Error json values'
+            )
+
+        for sec in section_request:
```
timlinux: sec -> section

django_project/lesson/views/section.py
ismailsunni: Ah yes. Thanks.

django_project/lesson/views/section.py
ismailsunni: section is already used, I changed it with `for section_request in sections_request`

PR #687, opened 2018-01-17
Internationalisation support for lesson app.

ismailsunni:
Ticker: #657 

lesson_translation

- Add new section will automatically add translation in the current language.
- Editing will only edit the translation, except English (it will also update the original value).

For @Gustry and @myarjunar, I have registered the fields for all models to be translated. But, you need to use `from modeltranslation.forms import TranslationModelForm` in your ModelForm.

### DIFF
--- a/deployment/docker/REQUIREMENTS.txt
+++ b/deployment/docker/REQUIREMENTS.txt
@@ -35,3 +35,4 @@ django-allauth==0.25.2
 Unidecode==0.4.20
 reportlab==3.4.0
 xhtml2pdf==0.2b1
+django-modeltranslation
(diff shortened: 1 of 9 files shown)


Gustry:
Thanks for the prototype @ismailsunni!

Some notes, in my head, I was expecting an HStore storing maybe. Having one field per language seems a lot. Not strong opinion. It depends of the user case.
And what happen if you update the indonesian field but not the English?
Also what happen if you add one english sentence in a big block of English? Users need to know that they need to update the French/Indonesian ones too.

Anyway, dealing with translations is not easy!

ismailsunni:
@timlinux I have fixed the migration conflict here.
@Gustry I have removed the section detail like in yours.

PR #696, opened 2018-01-21
Crud questions

Gustry:
This PR is adding CRUD for questions and answers.

Ticket #657

### DIFF
--- a/django_project/lesson/forms/question.py
+++ b/django_project/lesson/forms/question.py
@@ -1,0 +1,51 @@
+# coding=utf-8
+"""Question form."""
+
+from django.utils.translation import ugettext_lazy as _
+
+from crispy_forms.helper import FormHelper
+from crispy_forms.layout import (
+    Layout,
+    Fieldset,
+    Field,
+    Submit,
+)
+
+from modeltranslation.forms import TranslationModelForm
+from lesson.models.worksheet_question import WorksheetQuestion
+
+
+class QuestionForm(TranslationModelForm):
+    """Form for creating Question"""
+
+    class Meta:
+        model = WorksheetQuestion
+        fields = (
+            'question',
+            'question_image',
+        )
+
+    def __init__(self, *args, **kwargs):
+        self.worksheet = kwargs.pop('worksheet')
+        self.helper = FormHelper()
+        layout = Layout(
+            Fieldset(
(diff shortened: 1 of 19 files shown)


django_project/lesson/forms/question.py
```diff
@@ -0,0 +1,51 @@
...
+    Field,
+    Submit,
+)
+
+from modeltranslation.forms import TranslationModelForm
+from lesson.models.worksheet_question import WorksheetQuestion
+
+
+class QuestionForm(TranslationModelForm):
+    """Form for creating specification"""
```
timlinux: cut and paste

django_project/lesson/models/answer.py
```diff
@@ -27,6 +28,7 @@ class Answer(models.Model):
         help_text=_('Answer number.'),
```
timlinux: Can you make the help text descriptive - what is answer number for?

django_project/lesson/models/answer.py
```diff
@@ -56,5 +58,12 @@ class Meta:
         app_label = 'lesson'
         unique_together = ['answer_number', 'question']
```
timlinux: Do we need to do this? When first entering an answer initially we will either need to ensure that answers are given a unique sequence number or remove this constraint. I probably prefer to remove this constraint and implement sortable behaviour on the question list.

django_project/lesson/models/worksheet_question.py
```diff
@@ -30,9 +30,12 @@ class WorksheetQuestion(models.Model):
     )
 
     question_number = models.IntegerField(
-        help_text=_('Question number.'),
+        help_text=_(
+            'Used to order the questions for a lesson into the correct '
```
timlinux: Yeah something like this would be better for answer number....

django_project/lesson/models/answer.py
Gustry: Sorry, I fail to understand. Whatever if we do the sortable behavior or not, we can't have two answers having the same number and belonging to the same question.

>  I probably prefer to remove this constraint and implement sortable behaviour on the question list.

I'm missing something. We need to keep this uniqueness, especially if we do the sortable behavior, we won't have duplicates.

PR #700, opened 2018-01-22
Project color picker

myarjunar:
color-picker


Gustry:
Maybe you should remove the last migration about the lesson? It's now in develop

timlinux:
Looks good thanks @myarjunar

PR #703, opened 2018-01-22
add a dummy shift in DB to avoid Integrity Error

Gustry:
Ah, it seems the DB got me! Since my commit to add some constraints on the order number, the ordering is not working anymore.

If we could update the DB in a transaction (without integrity constraints), everything would be good because the JSON from the ajax request is correct according the DB. But because the `save` is done separately on each record in the DB, it's failing.  I checked the Django doc but I'm not sure we can disable integrity constraints by doing a bulk update command.

As we won't have a big lesson with thoussand of worksheets, I think it's fine to do this small dummy temporary shift in the DB. What do you think @timlinux ? Or should I disable my work on `unique_together`.

### DIFF
--- a/django_project/lesson/views/question.py
+++ b/django_project/lesson/views/question.py
@@ -354,9 +354,12 @@ class QuestionOrderSubmitView(
         try:
             questions_request = json.loads(specifications_json)
         except ValueError:
-            raise Http404(
-                'Error json values'
-            )
+            raise Http404('Error json values')
+
+        # Add dummy shift in the DB to avoid Integrity about unique_together
+        for question in questions:
+            question.question_number += len(questions_request)
+            question.save()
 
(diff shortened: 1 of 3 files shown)


timlinux:
@Gustry IIRC the django docs say you should avoid using len() on querysets as it normally does lazy binding of the objects in a queryset. As soon as you call len it will unpack all the objects. Rather use qs.count() which just does a sql count(*) operation which is much more efficient.

Gustry:
@timlinux The `len` is not the QuerySet, but on the JSON object from the AJAX request.

timlinux:
Ah sorry I didnt read your code carefully enough.

You can also wrap the method with a decorator so that all the changes get committed in one go - but for now your approach seems fine.

PR #706, opened 2018-01-23
Add worksheet order, improve form, fix order in worksheet form

Gustry:
* Add worksheet order
* improve form title
* fix order in worksheet form


timlinux:
LGTm thanks @Gustry

PR #707, opened 2018-01-23
DRY refactoring about new order in Lesson app

Gustry:
Small refactoring about new order in Lesson app

All our models should have the column named `sequence_number` so we can easily implement the ordering. (for a future app)


django_project/lesson/tools.py
```diff
@@ -0,0 +1,40 @@
+# coding=utf-8
```
timlinux: maybe utilities.py would be a nicer name?

PR #710, opened 2018-01-23
Worksheet detail

myarjunar:
worksheet-detail


Gustry:
It looks very good for now. I think you can merge and then we can review on our local ;-)
Thanks !

timlinux:
We dont need that check icon next to check your knowledge @myarjunar

timlinux:
Oh sorry ignore my above comment @myarjunar

PR #715, opened 2018-01-25
add permissions on lesson app, review mixin

Gustry:
I think a few things to check on Projecta how permissions are done.
I tried a non-admin account on changelog.qgis.org, I can see icons to remove. But then I can't get the webpage. It's confusing to display it if we have a 302 error after.

Can you tell me also why the mixin `LoginRequiredMixin` is enough in this view? To remove a project, you shouldn't need only a login, you need a staff login for instance `StaffuserRequiredMixin` or to be the project owner.
https://github.com/kartoza/projecta/blob/develop/django_project/base/views/project.py#L166

### DIFF
--- a/django_project/base/forms.py
+++ b/django_project/base/forms.py
@@ -73,6 +73,15 @@ class ProjectForm(forms.ModelForm):
             'moderation queue.')
     )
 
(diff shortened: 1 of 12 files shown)


timlinux:
LGTM

PR #716, opened 2018-01-25
Revert bootstrap

Gustry:
Sorry @timlinux 

Or should we fix these issues raised on gitter?

Arf, it seems you updated `docs/LessonsApp.graffle` in your squashed commit. I think you updated this file again on your side. Is it OK if you commit it again after that? Be sure to keep it when you save.


timlinux:
We can do it more simply with #717

PR #718, opened 2018-01-25
remove some useless code in Lesson app

Gustry:
Another round of ~~useless~~ unnecessary code @timlinux ;-)

Not finished, I will add some tests


timlinux:
down with useless code! LGTM thanks @Gustry

PR #731, opened 2018-01-30
fix permissions and add about link

### LINKED ISSUE #689: Unable to add changelog to project even with write permisions
# Problem
I have been granted access (manager) for a project but I get a 404 when I try to edit the project and also there is no option to add the changelog.

404 when editing the project
error

No add option for changelog
no_versions

# Solution
Make it happy

ann26:
fix #689 

This PR:
- [x] Fixed permissions for changelog manager to create, update and delete versions and category
- [x] Added about link in certifying organisation page
- [x] Also fixed permissions for certification manager to see action buttons in certifying organisation list

changelog managers can see create button and should be able to create, update and delete versions
screenshot from 2018-01-30 14-51-51

add clickable about link
screenshot from 2018-01-30 14-52-04

### DIFF
--- a/django_project/changes/views/category.py
+++ b/django_project/changes/views/category.py
@@ -20,7 +20,7 @@ from django.http import HttpResponseRedirect, Http404
 from django.db import IntegrityError
 from django.db.models import Q
 from django.core.exceptions import ValidationError
-from braces.views import LoginRequiredMixin, StaffuserRequiredMixin
+from braces.views import LoginRequiredMixin
 from ..models import Category, Version
 from ..forms import CategoryForm
 
--- a/django_project/changes/views/category.py
+++ b/django_project/changes/views/category.py
@@ -128,7 +128,7 @@ class JSONCategoryListView(CategoryMixin, JSONResponseMixin, ListView):
         return qs
 
 
-class CategoryListView(CategoryMixin, LoginRequiredMixin, ListView):
+class CategoryListView(LoginRequiredMixin, CategoryMixin, ListView):
     """List view for Category."""
     context_object_name = 'categories'
     template_name = 'category/list.html'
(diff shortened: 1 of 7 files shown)


django_project/changes/views/category.py
```diff
@@ -506,11 +510,16 @@ def get_queryset(self):
         projects which user created (staff gets all projects)
         :rtype: QuerySet
         """
+        self.project_slug = self.kwargs.get('project_slug', None)
```
Gustry: It's just my advice, but I try to not store all variables in `self`. If I don't store the variable in `self`, at least the next developer coming to read this code, he is aware that it's a local variable to this function and it will not be altered or read in another function. I try to avoid `self` whenever I can.

ann26: I see, ok I'll fix it

django_project/changes/views/category.py
```diff
@@ -127,7 +128,7 @@ def get_queryset(self):
         return qs
 
 
-class CategoryListView(CategoryMixin, StaffuserRequiredMixin, ListView):
+class CategoryListView(CategoryMixin, LoginRequiredMixin, ListView):
```
Gustry: Might good to follow this pattern: https://docs.djangoproject.com/en/2.0/topics/auth/default/#the-loginrequired-mixin

> This mixin should be at the leftmost position in the inheritance list.

https://stackoverflow.com/a/35746181/2395485

Maybe you can check them all?

Gustry: Might good to follow this pattern: https://docs.djangoproject.com/en/2.0/topics/auth/default/#the-loginrequired-mixin

> This mixin should be at the leftmost position in the inheritance list.

https://stackoverflow.com/a/35746181/2395485

Maybe you can check them all?

django_project/changes/views/category.py
```diff
@@ -599,7 +610,7 @@ def get_queryset(self):
         return self.queryset
 
 
-class ApproveCategoryView(CategoryMixin, StaffuserRequiredMixin, RedirectView):
+class ApproveCategoryView(CategoryMixin, LoginRequiredMixin, RedirectView):
```
Gustry: Same here

django_project/changes/views/category.py
ann26: ok @Gustry

timlinux:
LGTM thanks

PR #745, opened 2018-02-05
re order entries within the same category

Gustry:
Ticket #30 

order

I can still refactor the code later to avoid the copy/paste of the HTML template.

@timlinux @ann26 I need to check why I need to remove this variable from the context. I'm not sure about this middleware:
If I don't introduce this hack, I get:
```
Traceback:
File "/usr/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  157.                     response = middleware_method(request, response)
File "/home/web/django_project/core/custom_middleware.py" in process_template_response
  98.             context['the_version'] = context.get('entry').version

Exception Type: AttributeError at /en/1/version/35/order/1
Exception Value: 'QuerySet' object has no attribute 'version'
```

I have a list of entry L98, so it's normal that `version` doesn't work on a collection of objects.
https://github.com/kartoza/projecta/blob/develop/django_project/core/custom_middleware.py#L98

### DIFF
--- a/django_project/changes/views/entry.py
+++ b/django_project/changes/views/entry.py
@@ -20,8 +20,9 @@ from django.core.exceptions import ValidationError
 from django.http import HttpResponseRedirect
 from braces.views import LoginRequiredMixin, StaffuserRequiredMixin
 from pure_pagination.mixins import PaginationMixin
-from ..models import Version, Entry
+from ..models import Version, Entry, Category
 from ..forms import EntryForm
+from lesson.utilities import re_order_features
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


django_project/changes/views/entry.py
```diff
@@ -442,3 +443,74 @@ def get_redirect_url(self, pk):
...
+        :rtype: dict
+        """
+        context = super(EntryOrderView, self).get_context_data(**kwargs)
+        version_pk = self.kwargs.get('version_pk', None)
+        category_pk = self.kwargs.get('category_pk', None)
+        context['version'] = get_object_or_404(Version, pk=version_pk)
+        context['category'] = get_object_or_404(Category, pk=category_pk)
+
+        # Hack, the middleware is expecting a single object in this "entry"
+        del context['entry']
```
Gustry: Here @timlinux

Gustry:
My issue is fixed with the help of Tim.

Issue #740, opened 2018-02-03
We need to add support for defining curricula in the lessons app


timlinux:
See updated schema below which includes the curricula model.

![lessonsapp](https://user-images.githubusercontent.com/178003/35772817-5751f922-094d-11e8-81e1-5b95daff2816.png)

@charlotte-morgan please comment if there are specific attributes you would like to see in the curruciculum form / output report.

Gustry:
Between worksheet and curriculum:
* Each curriculum has zero or more worksheets but also each worksheet can be in zero or more curriculum. So it's becoming a many to many relationship (missing on the drawing)
* Each worksheet in a curriculum can have a different order in the curriculum, so we need another sequence number in this many to many relationship.

timlinux:
Thanks @Gustry - I have updated the diagram accordingly here:

![lessonsapp](https://user-images.githubusercontent.com/178003/35782465-36cf2620-0a01-11e8-993d-ca17e85db588.png)

ismailsunni:
- [ ] We need to implement the front end. :point_up: [February 25, 2018 5:45 AM](https://gitter.im/kartoza/projecta?at=5a91eaefe4ff28713ac11c34)

Gustry:
You need to implement everything, only models are there ;-)

ismailsunni:
Hehe, sure @Gustry 

@timlinux  Is there any discussion about how will the front end look like? If not, perhaps I will create some mock-up before coding

timlinux:
@ismailsunni yes please go ahead and make a mockup and then we can discuss it (and also show it to @charlotte-morgan).

haubourg:
@timlinux Hi Tim, 
if you plan to store personal data of students, please take attention to the European General Data Protection Regulation at https://www.eugdpr.org/ that instores privacy by design.
That may require some user features to allow data download or deletion. Data encryption and explicit user agreement are also mandatory. 
Even if you are hosted outside of EU, as soon as EU citizen are in the database, we are submitted to it.
Cheers Régis

timlinux:
@haubourg thanks for your comment. It does not really relate to the curriculum aspect (which details only how a collection of lessons can be created and named and printed out) so can I suggest to create a separate issue for this?

Also please comment in that ticket on whether law has any impact on us as a non-eu based project (QGIS is registered in Switzerland). We do plan to deploy under SSL regardless and having a mechanisme that course attendees can request their data to be removed sounds fine.

PR #750, opened 2018-02-06
add curriculum models

Gustry:
Add models for the curriculum
#740

### DIFF
--- a/django_project/lesson/models/curriculum.py
+++ b/django_project/lesson/models/curriculum.py
@@ -1,0 +1,102 @@
+# coding=utf-8
+
+"""Curriculum model definitions for lesson apps."""
+
+import logging
+import os
+from unidecode import unidecode
+
+from django.conf.global_settings import MEDIA_ROOT
+from django.contrib.auth.models import User
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+from django.utils.text import slugify
+
+from base.models.project import Project
+from core.settings.contrib import STOP_WORDS
+
+logger = logging.getLogger(__name__)
+
+
+class Curriculum(models.Model):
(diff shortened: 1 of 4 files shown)


django_project/lesson/models/curriculum.py
```diff
@@ -0,0 +1,102 @@
...
+from base.models.project import Project
+from core.settings.contrib import STOP_WORDS
+
+logger = logging.getLogger(__name__)
+
+
+class Curriculum(models.Model):
+    """Curriculum lesson model.
+
+    A curriculum is a grouping for one or more worksheets.
```
timlinux: owned by a specific user

django_project/lesson/models/curriculum_worksheets.py
```diff
@@ -0,0 +1,34 @@
...
+"""Curriculum model definitions for lesson apps."""
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from lesson.models.curriculum import Curriculum
+from lesson.models.worksheet import Worksheet
+
+
+class CurriculumWorksheets(models.Model):
```
timlinux: I wonder why we can't just use a manytomany field in the curriculum model. See https://docs.djangoproject.com/en/2.0/topics/db/examples/many_to_many/

Gustry: Correct me if I'm wrong, but we can't because we have the sequence number field on the relationship.
https://docs.djangoproject.com/en/dev/topics/db/models/#extra-fields-on-many-to-many-relationships

timlinux: Ah good point - LGTM then!

PR #752, opened 2018-02-07
alter description length to be 1000 characters

### LINKED ISSUE #669: description field is too short
the description field is super short. It would be good to allow a bit more of text so that a course can be explained

ann26:
fix #669 

This PR:
- [x] Changed the descriptions length in course type to be 1000 characters (previously 250)
screenshot from 2018-02-07 10-36-56

### DIFF
--- a/django_project/certification/models/course_type.py
+++ b/django_project/certification/models/course_type.py
@@ -41,7 +41,7 @@ class CourseType(models.Model):
 
     description = models.TextField(
         help_text=_('Course type description.'),
-        max_length=250,
+        max_length=1000,
         null=True,
         blank=True,
     )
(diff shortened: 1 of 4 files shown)


django_project/certification/models/course_type.py
```diff
@@ -41,7 +41,7 @@ class CourseType(models.Model):
 
     description = models.TextField(
         help_text=_('Course type description.'),
```
timlinux: Course type description - 1000 character limit

timlinux:
LGTM

django_project/certification/models/course_type.py
ann26: done @timlinux 
here is what look like in forms
screenshot from 2018-02-07 17-46-55

PR #754, opened 2018-02-07
fix uniquness of slug in Lesson app

Gustry:
We need to hit the database twice:
* to update the slug based on its current title/name
* include the ID after the first `save`

So now, it will update the slug on every update. Is-it fine to break permalink made before?


timlinux:
> So now, it will update the slug on every update. Is-it fine to break permalink made before?

Yes for now early in the development, it is no problem. Maybe we should only append the id if we see it is not unique? Otherwise LGTM

Gustry:
No the problem is not adding the ID at the end. I think we should do that always for newly created object.

My question about permalink is if you update the title/name of the section after 3 weeks time. The slug is created at the beginning. If you update the name of the section, do we expect to update slug too? (and to break all permalink made to this section)

PR #760, opened 2018-02-11
Added rule to ignore logging of invalid host errors

timlinux:


timlinux:
Note there is an error in the PR - replace null with 'logging:NullHandler'

PR #766, opened 2018-02-19
fix duplicate queries

### LINKED ISSUE #763: Duplicated Unapproved Certifying Orgs
There have been duplicate entries for some organizations showing up on the Unapproved Certifying Organisations page
[http://changelog.qgis.org/en/qgis/pending-certifyingorganisation/list/](url)

For example, there were duplicate CampToCamp and CampToCamp SA organizations.  
Deleting one CampToCamp entry removed the other. 
Approving one CampToCampSA entry resulting in one approved organization and removed both entries from the Unapproved Certifying Organisations page.
This behavior indicates that while showing up as duplicates they in fact represent single entries.

ann26:
fix #763 

This PR fixes duplicate pending certifying organisation queries when a user has multiple role within a project.


timlinux:
LGTM thanks @ann26

PR #773, opened 2018-02-20
use the `url` property instead of doing URL concatenation for media

Gustry:
In Pycharm, search and replace with regex:
From `\{\{ MEDIA_URL \}\}\{\{ (.*) \}\}` to `{{ $1.url }}`

I find this code better than a URL concatenation. I guess django is doing the URL smartly.

Following this blogpost http://staticfiles.productiondjango.com/blog/stop-using-static-url-in-templates/


timlinux:
LGTM thanks

PR #775, opened 2018-02-20
Allow only certifying organisation managers to access 'call to action'

### LINKED ISSUE #678: Remove call to action on cert page for users who are not part of the org
This is what I see when not logged in:

screen shot 2018-01-15 at 01 27 38

The buttons should not be shown anyone except cert org managers

sonlinux:
Fix #678 
Allow only certying organisation managers to access call to action

# non loged in user and non certying organisation managers will not see call to action as shown below.

call-to-action


timlinux:
LGTM thanks...

PR #780, opened 2018-02-21
recommend logo size

### LINKED ISSUE #531: We need to specify preferred size for logos in project view
specify logo image size else resize

It becomes a guessing game otherwise...

sonlinux:
Fix #531

I tested a few resolutions and sizes for the logo and a size of 512x512 with a resolution of 2048x2018 matches well. We can recommend these dimensions to be uploaded.

Appearance of the above dimensions

2048by2018


Gustry:
@sonlinux If you update the models, you need to do `make update-migration`. It will regenerate the migration file. But I think you should wait for you other PR about project models, because you migration will run into a conflict (or do this fix in the other PR so you can have one migration file)

sonlinux:
@Gustry I think I lost the migration change on a commit reset earlier on.
Let me fix

django_project/base/models/project.py
```diff
@@ -105,7 +105,9 @@ class Project(models.Model):
     image_file = models.ImageField(
         help_text=_('A logo image for this project. '
                     'Most browsers support dragging the image directly on to '
-                    'the "Choose File" button above.'),
+                    'the "Choose File" button above. '
+                    'We recommend that you upload a logo of size 512 x 512 pixels '
+                    'and 2048 x 2018 pixels resolution.'),
```
timlinux: This text does not make any sense for me. Size is the physical dimension of the image. Resolution is how many DPI (Dots Per Inch) it is designed to be displayed at. Typically print uses 300dpi (though it can vary) and screen on older tech was 96dpi, newer devices much higher. As it is now this text will just confuse people. Why don't we just say "The ideal size for your image should be X x Y."

I'm also confused by your image ratio you have 2048x2018 - is that a typo? 20148 and 20148 maybe?

sonlinux: Thanks @timlinux 
Am fixing that

django_project/base/migrations/0020_auto_20180222_1423.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('base', '0019_project_project_repository_url'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='project',
+            name='image_file',
+            field=models.ImageField(help_text='A logo image for this project. The ideal size for your image should be 512 x 512 pixels.', upload_to=b'images/projects', blank=True),
```
timlinux: what does the b do?

django_project/base/models/project.py
```diff
@@ -104,8 +104,8 @@ class Project(models.Model):
 
     image_file = models.ImageField(
         help_text=_('A logo image for this project. '
-                    'Most browsers support dragging the image directly on to '
-                    'the "Choose File" button above.'),
+                    'The ideal size for your image '
```
timlinux: Why not keep both text hints (for drag and drop and size hint)

should be -> is

sonlinux:
Thanks @timlinux 
I updated.

django_project/base/models/project.py
```diff
@@ -104,8 +104,9 @@ class Project(models.Model):
 
     image_file = models.ImageField(
         help_text=_('A logo image for this project. '
-                    'Most browsers support dragging the image directly on to '
-                    'the "Choose File" button above.'),
+                    'Most browsers support dragging the image '
+                    'directly on to the "Choose File" button above. '
+                    'The ideal size for your image is 512 x 512 pixels.'),
```
timlinux: It was a long road to get here, and one last step is to obey the 'pull left' rule. Please see:

http://kartoza.com/en/coding-standards/.  (search for pull left on that page)....

Gustry:
Thanks @sonlinux 
Even if it was not from you (the indentation), we should fix the code layout from previous developers ;-)

Did you setup your Pycharm to show you the right margin of 79 characters? In your editor, you should only split the sentence at the of the maximum 79 characters (without spliting one word in two pieces)

screen shot 2018-02-26 at 11 10 01

Check the line on the right

sonlinux:
Thanks @timlinux 
I will do the fix, 
No i have not yet set my pycharm to detect line length,
Am doing that just right away.

sonlinux:
Hi @Gustry
Can we merge this work?

Gustry:
I already approved 4 hours ago :-)

PR #784, opened 2018-02-22
fix redirection when not enough permissions to view a page

Gustry:
After adding/updating an entry in the changelog, a normal user will be redirected to the pending queue. We should add this method to redirect users when they don't have enough permissions to see the view.


timlinux:
LGTM thanks @Gustry - will you deploy to production too please?

Issue #723, opened 2018-01-25
Table made with markdown is not supported


Gustry:
From @christodewit in the Lesson app:

This is not displaying correctly in Projecta

Markdown in GitHub:
![screen shot 2018-01-25 at 13 24 06](https://user-images.githubusercontent.com/25034035/35386063-451f1f2c-01d3-11e8-968b-1b45b4607042.png)

Display in GitHub:
![screen shot 2018-01-25 at 13 24 15](https://user-images.githubusercontent.com/25034035/35386065-490ae8f0-01d3-11e8-9b74-d065fad3c91b.png)

Display in Projecta:
![screen shot 2018-01-25 at 13 27 56](https://user-images.githubusercontent.com/25034035/35386140-98339378-01d3-11e8-817c-5ef04af42cb7.png)

ThomasG77:
It's because it's not standard Markdown but GFM (Github Flavoured Markdown) See https://github.github.com/gfm/#tables-extension-

You will need to replace `markdown.markdown` https://github.com/kartoza/projecta/blob/536ed6c87b70fbd9e8be15ed48d345987630ba91/django_project/base/templatetags/custom_markup.py#L16 with another library supporting GFM otherwise, it's not possible to fix the issue.

timlinux:
Nice to have table support now - thanks @ThomasG77. @ismailsunni can you make them look beautiful now?

![screen shot 2018-02-27 at 16 26 17](https://user-images.githubusercontent.com/178003/36734080-27d12718-1bdb-11e8-8afc-08eb8a0c126f.png)


Ideally use some bootstrap magic to make them look more aesthetically pleasing....

Gustry:
Yes, I made a quick CSS to display them at least. I kept the ticket open to improve the CSS. @ismailsunni you can go ahead if you want

timlinux:
Before closing this, lets get nicer CSS for the table.

PR #785, opened 2018-02-23
Add support for tables in markdown using extensions

ThomasG77:
I've tried the suggested function in #723 and it works.
See screenshot

Although, the PR solves the GFM issue, the remaining job before closing completely the issue would be to style tables as there are no borders or styles applied (screenshot can be misleading as I used a third party dev tools to highlight table and cells in red)


timlinux:
Ah that is really cool - thanks @ThomasG77 

@Gustry can you add the needed styles and get this merged?

Gustry:
Ok, let's merge it. It's anyway better than before! ;D
So I will continue this PR with develop.

Merci Thomas!

PR #790, opened 2018-02-26
Small Improvement

ismailsunni:
Some small improvements (trying to jump into the projecta wagon again)

### DIFF
--- a/README-dev.md
+++ b/README-dev.md
@@ -37,6 +37,17 @@ make build-devweb
 make devweb
 ```
 
+In case you don't get some not installed packages, you can run this 
+repeatable steps:
+
+1. Close PyCharm
+2. `make kill` or `docker-compose -p projecta kill`
+3. `make rm` or `docker-compose -p projecta rm`
+4. `make build`
+5. `make build-devweb`
+6. `make devweb`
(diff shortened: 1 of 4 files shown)


README-dev.md
```diff
@@ -136,4 +136,16 @@ common cause of this is if you are running the server in both production
...
+ package when I run server in PyCharm.
+
+**A:** From Tim, run these steps:
+1. Close PyCharm
+2. `make kill` / `docker-compose -p projecta kill`
+3. `make rm` / `docker-compose -p projecta rm`
+4. `make build`
+5. `make build-devweb`
+6. `make devweb`
+7. start PyCharm
```
Gustry: Can you edit the main workflow? 
https://github.com/ismailsunni/projecta/blob/3ecf4faf7975766746b05aa963b077d862c8bfba/README-docker.md#production

Maybe add a new block which commands to do when the code is updated?

When I have an error, I don't think going to the FAQ. It's more like a bug for me and not a question. And I never been to this page.

make dbrestore, make migrate are missing from this list.

timlinux:
Can you add make status to the help section for make help? Its at the bottom of the makefile. Then LGTM

PR #791, opened 2018-02-26
remove Python warning about psycopg2

Gustry:
```
/usr/local/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning:
The psycopg2 wheel package will be renamed from release 2.8; 
in order to keep installing from binary please use "pip install psycopg2-binary" instead.
 For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
```

But from the website: https://pypi.python.org/pypi/psycopg2-binary
> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

So ?

### DIFF
--- a/deployment/docker/REQUIREMENTS.txt
+++ b/deployment/docker/REQUIREMENTS.txt
@@ -1,4 +1,4 @@
-psycopg2
+psycopg2-binary
 Django>=1.8,<1.9
 django-countries==3.4.1
 django-crispy-forms==1.4.0


timlinux:
Why do they advise to build from sources @Gustry ?

ismailsunni:
@timlinux @Gustry perhaps this can explain http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/

> Unfortunately, after the packages were released, it was reported of occasional segfaults of Python processes using Psycopg from the Wheel package. This was traced to the use of the Python ssl library concurrently with Psycopg opening connections, for instance in a multithread program opening concurrently https resources and database connections. The problem seems caused by a non-reentrant OpenSSL initialization function (unfortunately invoked by libpq at every connection) and the fact that the Python process ends up binding two different versions of the libssl library: the system one via the Python ssl library (e.g. imported by urllib or requests) and the Wheel one imported by the libpq, in turn imported by psycopg2.

> While the problem doesn't affect many users, a library behaving unreliably in combination with part of the stdlib is a situation less than optimal. The workaround is to force installing Psycopg from source, but this must be specified explicitly in the project dependencies (e.g. using the --no-binary flag in the pip command line or in the requirements.txt file); the Python packaging system doesn't really have a way to declare something like "install a package preferably from source"... so we had to create one ourselves.

It seems realiability issue if we use the binary one for production.

timlinux:
Does the source build require a c++ build chain in the container?

ismailsunni:
@timlinux 
It seems no, I try to install it using this http://initd.org/psycopg/docs/install.html#disabling-wheel-packages-for-psycopg-2-7, all is good.

Which way that we should use?

PR #793, opened 2018-02-26
remove unnecessary code, version slug is not unique so fetch it with the project too

Gustry:
Fix http://sentry.kartoza.com/sentry/projecta/issues/41/

* remove unnecessary code (already done by the django framework)
* don't use `self` everytime to store variable, it makes it harder to know which variable are global in `self` or local to a method
* fix fetching Version object only with its slug, use the project too.

### DIFF
--- a/django_project/changes/views/entry.py
+++ b/django_project/changes/views/entry.py
@@ -17,7 +17,6 @@ from django.views.generic import (
 )
 from django.db import IntegrityError
 from django.core.exceptions import ValidationError
-from django.http import HttpResponseRedirect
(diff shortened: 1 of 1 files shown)


timlinux:
@Gustry can you just confirm that when we remove get_context_data that other views e.g. list views will be properly filtered according to the current project and version?

LGTM thanks!

Gustry:
This `get_context_data` was in the `EntryCreateView`, so I fail to understand why we need all entries in the Entry creation form. I checked the template, there is no HTML using existing entries in this form

PR #795, opened 2018-02-26
fix multiple objects returned in version update view

Gustry:
fix multiple objects returned in version update view

Fix http://sentry.kartoza.com/sentry/projecta/issues/36/


timlinux:
LGTM thanks

PR #798, opened 2018-02-26
remove some unnecessary code in the Entry view

Gustry:


timlinux:
LGTM thanks @Gustry 

It would also be nice to have some view tests here like I asked of @sonlinux that verify the permissions behave correctly - that you cant delete items / up date items etc if you are not in the correct permissions group.

Gustry:
I will add tests after the PR from @sonlinux. He did some nice work so I will copy and adapt after he finished.

PR #800, opened 2018-02-27
Fix slug duplicates

### LINKED ISSUE #769: 500 error when editing organisation
This is an error encountered by Camptocamp. 
When they attempt to edit the Camptocamp SA organisation they receive a 500 error:

_Ooops Something broke. 
The powers that be have been informed.
In the meantime, please form an orderly queue and head to the home page.

If you need assistance, you may reference this error as b63834a31b1e46efba764837596ef5ae._

ann26:
fix #769 

- [x] Fixed problem when there is duplicates in certifying organisation slug.

Slug was created based upon project name and certifying organisation name, since they are unique together. However because project name and certifying organisation name are case sensitive and slug is made case insensitive, sometimes there could be duplicates. 
Example in project QGIS, we can create: Test organisation and test Organisation. Both will have same slug: qgis-test-organisation.

This PR change the slug creation using only certifying organisation name, and check if there is already same slug in the database for certifying organisation. When there are duplicates, the `check_slug` function will increment the slug.
So, for earlier example when we already have an object with slug `qgis-test-organisation`, the second object will have slug `1-qgis-test-organisation`

- [x] Fixed problem when in the existing database there is already two objects with the same slug.
When approving the certifying organisation, this will check if the object slug has duplicates in approved certifying organisation objects. When there are duplicates, it will update the slug with new incremented slug.

### DIFF
--- a/django_project/certification/admin.py
+++ b/django_project/certification/admin.py
@@ -122,6 +122,7 @@ class CertifyingOrganisationAdmin(admin.ModelAdmin):
     """Certifying organisation admin model."""
 
     filter_horizontal = ('organisation_owners',)
+    search_fields = ['name']
 
     def queryset(self, request):
         """Ensure we use the correct manager.
--- a/django_project/certification/models/certifying_organisation.py
+++ b/django_project/certification/models/certifying_organisation.py
@@ -16,6 +16,7 @@ from unidecode import unidecode
 from django.contrib.auth.models import User
 from django_countries.fields import CountryField
 import logging
+from certification.utilities import check_slug
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


Gustry:
Fixing http://sentry.kartoza.com/sentry/projecta/issues/30/
and this one too I guess ?
http://sentry.kartoza.com/sentry/projecta/issues/40/

ann26:
@Gustry yes for http://sentry.kartoza.com/sentry/projecta/issues/30/. 

for http://sentry.kartoza.com/sentry/projecta/issues/40/ , for the existing approved certifying organisation in the database, the problem might still persist because the slug remains the same. But there will not be another new object with the same slug after this.

Gustry:
So if there is a few organizations, maybe you can edit them in django admin (the duplicated one) to make them unique?

ann26:
yes I can do that @Gustry I will do that later.

Gustry:
Ok cool. So Sentry shouldn't get duplicated slugs anymore! and we can close both issues. Thanks!

I will deploy as to staging (and production maybe) soon, we made quite a few fixes these last day

PR #801, opened 2018-02-27
fix unicode characters when we send an email in the certification app

Gustry:
Fix http://sentry.kartoza.com/sentry/projecta/issues/39/
Fix http://sentry.kartoza.com/sentry/projecta/issues/29/


timlinux:
LGTM thanks

PR #805, opened 2018-03-01
Make sure the pending category query set give distinct values.

### LINKED ISSUE #803: Adding a category shows multiple duplicate entries in the pending queue
I just added a new catagory to the version app

When I went to the pending queue I saw this:

screen shot 2018-02-28 at 22 17 13

When I approved the first item, all the categories were removed from the pending list.

Please make sure we only take unique categories in the pending list.

ismailsunni:
Fix #803 

Reference : https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.distinct

> By default, a QuerySet will not eliminate duplicate rows. In practice, this is rarely a problem, because simple queries such as Blog.objects.all() don’t introduce the possibility of duplicate result rows. However, if your query spans multiple tables, it’s possible to get duplicate results when a QuerySet is evaluated. That’s when you’d use distinct().

### DIFF
--- a/django_project/changes/views/category.py
+++ b/django_project/changes/views/category.py
@@ -601,7 +601,8 @@ class PendingCategoryListView(
                 self.queryset = Category.unapproved_objects.filter(
                     Q(project=self.project) &
                     (Q(project__owner=self.request.user) |
-                     Q(project__changelog_managers=self.request.user)))
+                     Q(project__changelog_managers=self.request.user))
+                ).distinct()
                 return self.queryset
             else:
                 raise Http404(


Gustry:
I'm not the SQL expert, but `distinct` looks like we should double check the initial query (why 10 results and not 15 categories).
But it seems according to the django documentation that it's normal. Thanks @ismailsunni it looks good ;-)

timlinux:
LGTM - maybe want to check in the unapproved objects manager if there is an underlying issue there.

ismailsunni:
> LGTM - maybe want to check in the unapproved objects manager if there is an underlying issue there.

I have checked, unapproved objects manager looks good. It only happens when we use the OR operator for Q objects. *perhaps from different tables.

PR #812, opened 2018-03-06
register attendee and add it into course in the same screen

### LINKED ISSUE #556: Feature: register participant and add using checkbox in the same screen
It would be easier to reduce the amount of steps to add the participants to the course.
In the screen where we add the participant to the list, it should be possible to check a box to automatically add the participant to the course.

ann26:
fix #556 

Add new attendee and automatically add the newly created attendee into course:
peek 2018-03-06 10-50

Only add new attendee:
peek 2018-03-06 10-52

Also changed icons in course details, because there are two identical icons for different purpose:
one was for uploading new attendee using csv file and the other one for paying unpaid certificate
screenshot from 2018-03-06 10-54-43

I changed it into this:
screenshot from 2018-03-06 11-04-59

### DIFF
--- a/django_project/certification/views/attendee.py
+++ b/django_project/certification/views/attendee.py
@@ -2,7 +2,6 @@
 import csv
 from django.db import transaction
 from django.core.urlresolvers import reverse
-from django.http import HttpResponseRedirect
 from django.views.generic import (
     CreateView, FormView)
 from braces.views import LoginRequiredMixin, FormMessagesMixin
--- a/django_project/certification/views/attendee.py
+++ b/django_project/certification/views/attendee.py
@@ -2,6 +2,7 @@
 import csv
 from django.db import transaction
 from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect
 from django.views.generic import (
     CreateView, FormView)
 from braces.views import LoginRequiredMixin, FormMessagesMixin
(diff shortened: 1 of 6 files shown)


timlinux:
Lovely thanks @ann26 LGTM

timlinux:
Ok can you please add view tests whenever you change the behaviour of a view....

Thanks!

ann26:
ok @timlinux , I will add the test view into this PR

ann26:
Hi @timlinux I added the view test. :)

django_project/certification/views/attendee.py
```diff
@@ -73,6 +82,24 @@ def get_form_kwargs(self):
...
+                object = form.save()
+                course_slug = self.kwargs.get('slug', None)
+                course = Course.objects.get(slug=course_slug)
+                course_attendee = CourseAttendee(
+                    attendee=object,
+                    course=course,
+                    author=self.request.user
+                )
+                course_attendee.save()
+        return HttpResponseRedirect(self.get_success_url())
```
Gustry: Just curious, why not letting the django framework do the redirection by itself after the `form_valid` method?

Gustry: Also no need to call the parent `form_valid`? I'm not sure, just asking

ann26: Because the redirection is based upon the `add_to_course` input. If `add_to_course` is checked, after `form_valid` it will create new course attendee using the newly created attendee object and then it will redirect to `course detail view`. But when `add_to_course` is not checked, after saving new attendee it will redirect to `course attendee create view`.

ann26: @Gustry , I think by calling form_valid, we make sure that it redirect after form validation and after saving the new objects. cmiiw.

Gustry: Well, if I read your code in this `form_valid`, it will, in any case, hit this return statement L101. Django framework will call by itself the `get_success_url` anyway if you call `return super(YourClassName, self).form_valid(form)` and Django will perform additional task included in the parent method.

ann26: I see... I have fixed it using `super(AttendeeCreateView, self).form_valid(form)` now. Thank you @Gustry :)

Gustry:
Thanks @ann26 I think it's better like this when we overwrite and then call the parent method ;-)

PR #816, opened 2018-03-06
fix permission with clean checks

### LINKED ISSUE #807: Cannot add course type, conveners and training centre
When logged in ad owner, I can *edit* any existing thing and add new courses but cannot add all the rest. the star button is never shown
image the screenshot is while hovering the Course types box
I used to be able to do this or I'm I missing something?

sonlinux:
fix #807 
Logged in as an owner you are now able to CRUD training centre, course type as well as course converners too.

### DIFF
--- a/django_project/certification/views/certifying_organisation.py
+++ b/django_project/certification/views/certifying_organisation.py
@@ -226,9 +226,8 @@ class CertifyingOrganisationDetailView(
             context['user_can_create'] = True
             context['user_can_delete'] = True
 
-        certification_managers = (
-            context['the_project'].certification_managers.all())
-        if self.request.user in certification_managers:
+        if context['the_project'].certification_managers.filter(
+                project__owner=self.request.user).exists():
             context['user_can_create'] = True
             context['user_can_delete'] = True
 
@@ -236,8 +235,8 @@ class CertifyingOrganisationDetailView(
             context['user_can_create'] = True
             context['user_can_delete'] = True
 
-        if self.request.user in \
-                certifying_organisation.organisation_owners.all():
+        if certifying_organisation.organisation_owners.filter(
+                project__owner=self.request.user).exists():
             context['user_can_create'] = True
             context['user_can_delete'] = True
         return context
--- a/django_project/certification/views/certifying_organisation.py
+++ b/django_project/certification/views/certifying_organisation.py
@@ -227,7 +227,7 @@ class CertifyingOrganisationDetailView(
             context['user_can_delete'] = True
 
         if context['the_project'].certification_managers.filter(
-                project__owner=self.request.user).exists():
+                project__owner=self.request.user.id).exists():
             context['user_can_create'] = True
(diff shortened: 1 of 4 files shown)


sonlinux:
Thanks @Gustry 
Checking that

sonlinux:
@Gustry the certification managers permissions seem okay and working as espected.
Unless there is something in particular we need to work on or fix,
Let me know.
Thanks

django_project/certification/views/certifying_organisation.py
```diff
@@ -217,6 +221,25 @@ def get_context_data(self, **kwargs):
...
             context['the_project'] = Project.objects.get(slug=project_slug)
             context['project'] = context['the_project']
+
+        if self.request.user.is_staff:
+            context['user_can_create'] = True
+            context['user_can_delete'] = True
+
+        certification_managers = (
+            context['the_project'].certification_managers.all())
+        if self.request.user in certification_managers:
```
ismailsunni: I just stumbled across this article, http://bookofstranger.com/optimizing-django-orm-queries-for-best-performance/

In that article, it's better to use `.exists()`. 

Another reference : https://docs.djangoproject.com/en/dev/ref/models/querysets/#exists

sonlinux: Thanks @ismailsunni 
ill refactor that.

Gustry: Will you change also in the previous refactoring you did? And in the Lesson app in the worksheet view. Thanks @ismailsunni and @sonlinux

sonlinux: Alright @Gustry 
Doing that just now

sonlinux: I updated this the queryset @ismailsunni and @Gustry

sonlinux:
Thanks @ismailsunni

django_project/certification/views/certifying_organisation.py
```diff
@@ -217,6 +221,24 @@ def get_context_data(self, **kwargs):
         if project_slug:
             context['the_project'] = Project.objects.get(slug=project_slug)
             context['project'] = context['the_project']
+
+        if self.request.user.is_staff:
+            context['user_can_create'] = True
+            context['user_can_delete'] = True
+
+        if context['the_project'].certification_managers.filter(
```
Gustry: L222, `the_project` is created only it the condition is true. Here line L229, you use it, but it will raise a nexception of "unknown variable" if "the_project" doesn't exist. I'm not sure, but I think we should be able to remove the "if" line 221, is-it possible to have the project_slug?

sonlinux: I think we can use the project_slug, let me test that.

sonlinux: @Gustry I think we can go with it without the contion on L221.
I have updated.

PR #817, opened 2018-03-06
add caret to language dropdown

sonlinux:
Add caret to language dropdown.

gustry


sonlinux:
Requested and approved by @Gustry  from gitter channel.

PR #827, opened 2018-03-09
fix changelog entry detail permission

### LINKED ISSUE #826: 'Add new entry' button not visible on version detail.
## Problem 
Add new entry button is missing (can not be seen) on version detail page.

## Solution
Fix permission to view the button

add_new_entry

sonlinux:
fix #826 

We needed to fix our new permission checks on the changelog entry detail too.
## logged in as owner

fix_version

## not logged in
unloged

### DIFF
--- a/django_project/changes/templates/version/detail.html
+++ b/django_project/changes/templates/version/detail.html
@@ -21,6 +21,7 @@
     <div class="row" style="margin-top:15px;margin-bottom:15px;">
         <div class="col-lg-12">
             <div class="btn-group btn-group pull-left">
+                {% if user_can_edit %}
                     <a class="btn btn-default btn-mini tooltip-toggle"
                        data-title="Download sponsor list as html"
                        href='{% url 'version-sponsor-download' project_slug=version.project.slug slug=version.slug %}'>
@@ -33,6 +34,7 @@
                         <span class="glyphicon glyphicon-download"></span>
                         <span style="font-size: 7pt;">R S T</span>
                     </a>
+                {% endif %}
                 <a class="btn btn-default btn-mini tooltip-toggle"
                    data-title="Download as GNU Changelog" data-toggle="tooltip"
                    href='{% url 'version-download-gnu' project_slug=version.project.slug slug=version.slug %}'>
--- a/django_project/changes/templates/version/detail.html
+++ b/django_project/changes/templates/version/detail.html
@@ -21,7 +21,6 @@
     <div class="row" style="margin-top:15px;margin-bottom:15px;">
         <div class="col-lg-12">
             <div class="btn-group btn-group pull-left">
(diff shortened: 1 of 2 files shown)


django_project/changes/templates/version/detail.html
```diff
@@ -34,7 +33,6 @@ <h1>Entries (all)</h1>
                         <span class="glyphicon glyphicon-download"></span>
                         <span style="font-size: 7pt;">R S T</span>
                     </a>
-                {% endif %}
```
Gustry: Are you sure to remove @sonlinux ?
Anonymous users mustn't be able to see these buttons I think

django_project/changes/views/version.py
```diff
@@ -175,6 +175,23 @@ def get_context_data(self, **kwargs):
...
+            context['user_can_delete'] = True
+
+        if context['project'].changelog_managers.filter(
+                project__owner=self.request.user.id).exists():
+            context['user_can_edit'] = True
+            context['user_can_delete'] = True
+
+        if self.request.user == context['project'].owner:
+            context['user_can_edit'] = True
+            context['user_can_delete'] = True
```
Gustry: I'm failing to understand, where this could be before? We were using these variables in the template, but not creating these variables in the view? Weird

django_project/changes/templates/version/detail.html
sonlinux: Alright @Gustry 
I was following the permission checks in master branch.
Let me do so

django_project/changes/views/version.py
sonlinux: Thanks
Fixed that initialisation.

django_project/changes/views/version.py
```diff
@@ -170,11 +170,30 @@ def get_context_data(self, **kwargs):
...
         context['sponsors'] = sponsors
+        context['user_can_edit'] = False
+        context['user_can_delete'] = False
         project_slug = self.kwargs.get('project_slug', None)
         context['project_slug'] = project_slug
         if project_slug:
             context['project'] = Project.objects.get(slug=project_slug)
 
+        # lets check for specific user permissions here.
+        if self.request.user.is_staff:
```
ismailsunni: If it's me, I would use variables:
is_staff = self.request.user.is_staff
is_superuser = self.request.user.is_superuser
and so on, then put them in one if. Since all of them has the same context.

ismailsunni: Or put all of them in one `if` (without temp variable), I think python will do it better optimization.
Or just use `elif` so that we don't evaluate all of them. *I like this one most.

sonlinux: Thanks 
Ill optimise when in the next PR for managing them in single file.

PR #834, opened 2018-03-10
Update table style using Boostrap & a custom class for header

ThomasG77:
Due to #723 follow-up

For reviewing, see the options I've choose to use at https://getbootstrap.com/docs/4.0/content/tables/, in particular for "Striped rows" (maybe you may not want)

The transformation from markdown to table generate two `<tbody>` instead of a `<theader>` & a `<tbody>`. It's the why I've made a change in the CSS file too.

### DIFF
--- a/django_project/base/templatetags/custom_markup.py
+++ b/django_project/base/templatetags/custom_markup.py
@@ -18,7 +18,7 @@ def base_markdown(value):
         safe_mode=True,
         enable_attributes=False)
     html_output = html_output.replace(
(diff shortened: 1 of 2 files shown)


Gustry:
Thanks @ThomasG77 for your contribution again.

This CSS is also used in the PDF rendering, in which the bootstrap library is not working so well, I think I even disabled it. I would need to check.

timlinux:
Thanks very much @ThomasG77 ! We will look at the PDF rendering separately

PR #839, opened 2018-03-12
Fix certifying orgnaisation owner permissions

### LINKED ISSUE #829: Edit options missing for certifying organisations
@sonlinux we have had some regressions in the certification platform that need to be fixed urgently.

The organisation owner is no longer able to edit course types and other content on  the certification organiation page e.g.: 

http://staging.changelog.qgis.org/en/qgis/certifyingorganisation/ihe-delft-institute-for-water-education/

There are also some strange things in the code like this can_delete sentinel which actually also provides edit access. This should be made logically consistent e.g.:

``projecta/django_project/certification/templates/certifying_organisation/detail.html``

```
        {% if user_can_delete %}
        <td style="padding-right: 0">
             <div class="btn-group pull-right button-action" style="display: none;">
                <a class="btn btn-default btn-xs btn-delete tooltip-toggle"
                   href='{% url "course-delete" project_slug=certifyingorganisation.project.slug organisation_slug=certifyingorganisation.slug slug=course.slug %}'
                   data-title="Delete">
                    <span class="glyphicon glyphicon-minus"></span>
                </a>
                <a class="btn btn-default btn-xs tooltip-toggle"
                   href='{% url "course-update" project_slug=certifyingorganisation.project.slug organisation_slug=certifyingorganisation.slug slug=course.slug %}'
                   data-title="Edit">
                    <span class="glyphicon glyphicon-pencil"></span>
                </a>
                 <a class="btn btn-default btn-xs btn-view tooltip-toggle"
                    href='{% url 'course-detail' project_slug=certifyingorganisation.project.slug organisation_slug=certifyingorganisation.slug slug=course.slug %}'
                    data-title="View">
                    <span class="glyphicon glyphicon-eye-open"></span>
                </a>
            </div>
        </td>
        {% elif user == course.course_convener.user %}
```

@sonlinux please fix this ASAP - people are trying to use the system and everything needs to work properly.

See also #667

sonlinux:


sonlinux:
Hi @guys the permission problem was brought about due to the change in querrying for the permissions to insing filter.
I have reverted to querrying all permissions and validating from that querryset for now.
I suggest we proceed like this until we move to using a single file for managing these checks.
Thanks

Gustry:
> I have reverted to querrying all permissions and validating from that querryset for now.

What was the issue? Any error? No need to fix the other ones too? ( Lesson app)

sonlinux:
The issue was the permissions on certifying organisation detail page where not working for certifying organisation managers.
I have fixed

sonlinux:
> No need to fix the other ones too? ( Lesson app)

Let me check and test the other ones first

sonlinux:
Its all okay now @Gustry

django_project/certification/templates/certifying_organisation/detail.html
```diff
@@ -268,7 +268,7 @@ <h5>No courses are defined, but you can <a
             {% endif %}
         </td>
 
-        {% if user_can_delete %}
+        {% if user_can_delete and user_can_create %}
```
Gustry: I'm failing to understand this one. It's for a "delete" button, why do you need the create too?

sonlinux: We have 2 actions in that div.
For delete as well as for creating.

Gustry: You need to move the if for each individual button maybe?

sonlinux: >You need to move the if for each individual button maybe?

We could but it works alright still like this as well as the early permission checks were implemented like that as well. 

It works alright for the delete check only,
I have cleaned it up.

sonlinux: I think it makes sense to as he who can delete means can also update or create.

Gustry:
Ok, we can definitely merge this one ;-) Keep the ticket open from Tim and add a comment. Thanks !

sonlinux:
Great thanks @Gustry

PR #844, opened 2018-03-12
Add markdown support.

### LINKED ISSUE #828: Lesson table should support markdown in cells
I would like to be able to bold certain words or otherwise use Markdown to embellish content in the specification table.

screen shot 2018-03-09 at 16 27 56

For example I want to make the word 'Algorithm' in the above image bold by doing ``**Algorithm**``.

Please run table text through the markdown filter.

sonlinux:
fix #828 
Add markdown support to  requirements table content.

mardown_suuport


Gustry:
Can you update the model too saying that markdown is supported. Thanks. There are already quite a few fields with this information.

sonlinux:
Alright @Gusty
I'll do so,
Thanks.

Issue #833, opened 2018-03-10
Lessons app attachment should use a more logical path


timlinux:
Currently when you attache worksheet data to the lesson app they go in the media folder under e.g.

```
images/lesson/worksheet/external_data/1c762fd1b2772b878dea1873af21e96bd9be71a0.zip
```

These files are not images and should be stored in a more logical path e.g.:

```
lesson/worksheet/external_data/1c762fd1b2772b878dea1873af21e96bd9be71a0.zip
```

timlinux:
Make sure that there is a plan to move existing data into the correct path too.

PR #845, opened 2018-03-12
Add link to sample data

### LINKED ISSUE #832: Lesson should provide a link to the attachment if provided
When adding an attachement to a lesson, there should be a link shown under the worksheet like this:

Click [here](#) to download the sample data for the lesson.

This should be shown on the bottom of the lesson page e.g.

http://changelog.qgis.org/en/qgis/section/composer-25/detail/124/

sonlinux:
fix #832 
sample_data

### DIFF
--- a/django_project/lesson/templates/worksheet/detail.html
+++ b/django_project/lesson/templates/worksheet/detail.html
@@ -341,6 +341,15 @@
             </ul>
         </div>
     </div>
+    <div class="row">
+        <div class="col-lg-12 text-center" style="margin-bottom: 10px;">
+            {% if worksheet.external_data %}
+            <p>Click <a href="{{ worksheet.external_data.url }}">here</a> to
+                download the sample data for
+                the lesson.</p>
+            {% endif %}
+        </div>
+    </div>
 
     <script>
         $('.btn-delete').hover(


Gustry:
As mentioned in the ticket, don't forget the PDF template too

sonlinux:
Hi @Gustry 
Yes I am testing with that just now.

sonlinux:
Do we need to show the download link in PDF @Gustry ?

ismailsunni:
I think it's good to have in the PDF also since we can click the link to the PDF.

Gustry:
:point_up: [March 12, 2018 3:12 PM](https://gitter.im/kartoza/projecta?at=5aa68ace0a1614b7120ef2df)

Raw clickable link in the PDF, so we can click and read the link

sonlinux:
Alright @Gustry 
Adding one

sonlinux:
Updated with PDF link @Gustry 

download_link

Gustry:
@sonlinux As I said, it needs to be "readable", if you print the PDF on a piece paper, we need to be able to read it

django_project/lesson/templates/worksheet/detail.html
```diff
@@ -341,6 +341,15 @@ <h3 class="custom-heading" style="margin-left:-40px"><strong>Further reading: </
             </ul>
         </div>
     </div>
+    <div class="row">
+        <div class="col-lg-12 text-center" style="margin-bottom: 10px;">
+            {% if worksheet.external_data %}
+            <p>Click <a href="{{ worksheet.external_data.url }}">here</a> to
```
Gustry: Make the link opening in a new tab, to avoid losing your current tab

django_project/lesson/templates/worksheet/print.html
```diff
@@ -196,6 +196,14 @@ <h2>Further reading:</h2>
     </ul>
 </section>
 <footer>
+     <div class="col-lg-12" style="margin-bottom: 10px;
```
Gustry: There is no bootstrap in that page. Please remove it. Look a few lines above for the further reading.

django_project/lesson/templates/worksheet/detail.html
sonlinux: When you click it only downloads the file and the current tab remains unchanged

django_project/lesson/templates/worksheet/print.html
```diff
@@ -196,6 +196,14 @@ <h2>Further reading:</h2>
...
+     padding-left: 100pt;">
+            {% if worksheet.external_data %}
+                <p>Click <a href="{{ request.META.HTTP_HOST }}{{ worksheet.external_data.url }}">here</a> to
```
Gustry: DIsplay the full link instead of "click" in case the PDF is printed on paper. Also open it in a new tab.

django_project/lesson/templates/worksheet/detail.html
Gustry: Hum, maybe you are right. ok so ;-)

django_project/lesson/templates/worksheet/print.html
sonlinux: Will it not be too long @Gustry 
since the link holds the filepath.

sonlinux: for instance
```
http://0.0.0.0:61202/media/images/lesson/worksheet/external_data/9792e82e2d69dd693ba79c405ae2323c21881284.dmp
```

sonlinux: Because when the file gets stored it gets to have that long unique name hence when displaying that it will be very long.
That issue needs to be resolved in 
https://github.com/kartoza/projecta/issues/833

Gustry: Right.
But these links are at end of the PDF anyway, at the bottom, and the link is clickable, right? So I don't see it as a blocker. Nobody will read it if they don't need.

I know for a training, we will come with our USB sticks etc with data inside. But if somehow, the PDF is printed, and the student lost/doesn't have the data, at least links are useful he can still manage. The best would to create a short link, but this it out of scope for now.

sonlinux: For openening in new tabs 
I have updated @Gustry

sonlinux:
point_up March 12, 2018 3:12 PM

PR #857, opened 2018-03-14
If the user is only manager, staff permission is not enough to display that page

Gustry:
If the user is only manager, staff permission is not enough to display that page

### DIFF
--- a/django_project/lesson/views/question.py
+++ b/django_project/lesson/views/question.py
@@ -11,7 +11,7 @@ from django.views.generic import (
 from django.shortcuts import get_object_or_404
 from django.utils.translation import ugettext_lazy as _
 
-from braces.views import LoginRequiredMixin, StaffuserRequiredMixin
+from braces.views import LoginRequiredMixin
 
 from lesson.forms.question import QuestionForm
 from lesson.models.worksheet_question import WorksheetQuestion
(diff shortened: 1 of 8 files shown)


django_project/lesson/views/question.py
```diff
@@ -126,8 +125,7 @@ def get_success_url(self):
         })
 
 
-class QuestionOrderView(
-    StaffuserRequiredMixin, QuestionMixin, ListView):
+class QuestionOrderView(LoginRequiredMixin, QuestionMixin, ListView):
     """List view to order questions"""
```
sonlinux: The code looks okay,
Lets add a dot **.** to end the docstring

django_project/lesson/views/section.py
```diff
@@ -175,7 +187,7 @@ def get_success_url(self):
         return url
 
 
-class SectionOrderView(StaffuserRequiredMixin, SectionMixin, ListView):
+class SectionOrderView(LoginRequiredMixin, SectionMixin, ListView):
     """List view to order section"""
```
sonlinux: Same here a **.** in the docstring

django_project/lesson/views/specification.py
```diff
@@ -127,7 +127,7 @@ def get_success_url(self):
 
 
 class SpecificationOrderView(
-    StaffuserRequiredMixin, SpecificationMixin, ListView):
+    LoginRequiredMixin, SpecificationMixin, ListView):
     """List view to order specifications"""
```
sonlinux: Same here a **.** in the docstring

django_project/lesson/views/question.py
Gustry: I will fix them in one go after this PR with a regex ;-)
Thanks

ann26:
Please fix the travis error :) otherwise LGTM

PR #858, opened 2018-03-14
fix missing markdown in PDF rendering

Gustry:
@sonlinux Follow up after https://github.com/kartoza/projecta/pull/844


sonlinux:
What fix is needed here @Gustry ?

Gustry:
You added the markdown support in the HTML template in #844, but you didn't add it in the PDF rendering. So the raw markdown will be displayed in the PDF.

sonlinux:
Ah I see, thanks @Gustry 
I didnt take note of the PDF template on time.

PR #861, opened 2018-03-15
Update i18n

### LINKED ISSUE #821: i18n Supports.
When I tried to print the PDF in Bahasa Indonesia, e.g. http://changelog.qgis.org/id/qgis/section/introduction-qgis-1/print/77/

I got these parts are not translated.
screen shot 2018-03-08 at 16 30 18

The other pars, `More about` is translated because it's in the model.

I think we should start making projecta support i18n (not only for the lesson app), but also the text.

cc @timlinux

ismailsunni:
Fix #821 

Please see also: https://github.com/kartoza/projecta/wiki/Internationalisation

screen shot 2018-03-15 at 22 50 19

The translation is working, but we still need to translate it (and mark the string that needs to be translated).


deployment/Makefile
```diff
@@ -152,7 +152,7 @@ shell:
 	@echo "------------------------------------------------------------------"
 	@echo "Shelling in in production mode"
 	@echo "------------------------------------------------------------------"
-	@docker-compose -p $(PROJECT_ID) exec uwsgi /bin/bash
+	@docker-compose -p $(PROJECT_ID) exec devweb /bin/bash
```
Gustry: This looks wrong. L153, you must be in production mode.

Maybe better to create another make command for dev mode?

deployment/Makefile
```diff
@@ -329,6 +329,35 @@ flake8:
 	@echo "Check flake 8"
 	@flake8 ../django_project
 
+update-translation-strings:
+	@echo
+	@echo "------------------------------------------------------------------"
+	@echo "Update string for translation."
```
Gustry: strings? plural?

deployment/Makefile
```diff
@@ -329,6 +329,35 @@ flake8:
...
+	@echo "------------------------------------------------------------------"
+	@docker-compose -p $(PROJECT_ID) exec uwsgi python manage.py makemessages
+
+compile-translation-strings:
+	@echo
+	@echo "------------------------------------------------------------------"
+	@echo "Compile string for translation."
```
Gustry: same here

deployment/Makefile
```diff
@@ -329,6 +329,35 @@ flake8:
...
+	@echo "------------------------------------------------------------------"
+	@docker-compose -p $(PROJECT_ID) exec uwsgi python manage.py compilemessages
+
+push-translation-source:
+	@echo
+	@echo "------------------------------------------------------------------"
+	@echo "Push translation source and translation to transifex."
```
Gustry: This help looks weird. I would avoid the second word 'translation'. Maybe just "push translations to transifex"

deployment/Makefile
```diff
@@ -329,6 +329,35 @@ flake8:
...
+	@echo "You need to set up your transifex client first."
+	@echo "------------------------------------------------------------------"
+	@tx push -s -t
+
+pull-translation-source:
+	@echo
+	@echo "------------------------------------------------------------------"
+	@echo "Pull all translation source and translation from transifex."
```
Gustry: Same here.

deployment/Makefile
```diff
@@ -371,4 +400,7 @@ help:
 	@echo "* **update-migrations** - freshen all migration definitions to match the current code base."
 	@echo "* **web** - same as **run** - runs the production site."
 	@echo "* **pep8** - Run Python PEP8 check."
-
+	@echo "* **update-translation-strings** - Update string for translation."
```
Gustry: Plural too I think

deployment/Makefile
ismailsunni: ah thanks, perhaps I was too sleepy

ismailsunni:
> You seem to forget the PDF template, the most important for your training :-)

I didn't forget but just too exhausted last night. I will do for the print also for sure. Thanks for reminding me.

deployment/Makefile
ismailsunni: Ah yes, I replace it incidentally. That's why we need peer review :). Thanks a lot.

ismailsunni:
I need this for the training :), I will merge and deploy to staging for testing.

PR #865, opened 2018-03-20
Fix more about image error

### LINKED ISSUE #837: clicking on the 'more about' image to see a larger version does not work
we need this functionality on the main exercise image as well, potentially all images.

sonlinux:
fix #837 

- [x] You can now preview as well as click to view original image in new taab on worksheet detail.

worksheet

### DIFF
--- a/django_project/lesson/templates/worksheet/detail.html
+++ b/django_project/lesson/templates/worksheet/detail.html
@@ -180,7 +180,7 @@
         </div>
         {% if worksheet.exercise_image %}
             <a href="#" class="pop-image">
-                <img id="/media/{{ worksheet.exercise_image }}" class="img-responsive img-rounded pull-right"
+                <img id="{{ worksheet.more_about_image.url }}" class="img-responsive img-rounded pull-right"
                      src="{{ worksheet.exercise_image.url }}"
                      style="max-width: 100% !important;"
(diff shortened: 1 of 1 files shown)


django_project/lesson/templates/worksheet/detail.html
```diff
@@ -180,7 +180,7 @@ <h4><strong>
         </div>
         {% if worksheet.exercise_image %}
             <a href="#" class="pop-image">
-                <img id="{{ worksheet.exercise_image }}" class="img-responsive img-rounded pull-right"
+                <img id="/media/{{ worksheet.exercise_image }}" class="img-responsive img-rounded pull-right"
```
Gustry: Sorry, I still have notifications from some repositories ;-)
What about `{{ worksheet.exercise_image.url }}` so if you don't use `media/` anymore one day, nothing is broken.

sonlinux:
:-) good to hear from u @Gustry 
Let me do so

gubuntu:
has this been tested and can it be deployed today still (before tomorrow's training)?

timlinux:
> has this been tested and can it be deployed today still (before tomorrow's training)?

Arent they working from PDF's at this point? I think its too late to roll anything out for the training....

PR #872, opened 2018-03-28
fix action buttons layout on lesson question.

### LINKED ISSUE #841: Adding a question squashes the question edits into a single row
In the lessons app if I put the second question, the edit and remove widgets are squashed into one row.

screen shot 2018-03-12 at 13 39 55

sonlinux:
fix #841 

- [x] When more than one question is added the action buttons lose their alignemt and get squashed in a row on the left (Fixed)

button_fix


timlinux:
Thanks for this @sonlinux

PR #873, opened 2018-03-29
project approved by default

### LINKED ISSUE #736: No way to see approval requests for new projects
When a new project is registered, there is no convenient link for an admin to see. Maybe we should just approve it by default.....

ann26:
fix #736 

Project is approved by default and redirect user to home after creation.
peek 2018-03-30 05-55


timlinux:
Great thanks!

PR #874, opened 2018-04-02
fix print bug and convener slug duplicates

### LINKED ISSUE #868: Error 500 when trying to print a certificate
This error produces when I try to print a new certificate or regenerate it.

If you need assistance, you may reference this error as e5bcad001ba1496c9f4bcf447afea2f9.

ann26:
fix #868 

This PR fixes:
- Error when printing using special characters
- Duplicate slug for course convener

Printed certificate that initially fails:
screenshot from 2018-04-02 09-32-10


timlinux:
LGTM thanks @ann26 - can you deploy it live once merged?

PR #877, opened 2018-04-05
Fix missing image

ismailsunni:
- Fix missing image URL in the template
- Add codecov badge

The bug was in this url : http://changelog.kartoza.com/en/inasafe/section/introduction-to-inasafe/detail/11/


ann26:
LGTM

PR #881, opened 2018-04-05
tweak domain

ann26:
- Made `changelog.kartoza.com` as default site: where we can register domain.
- If we want to only show projects that belongs to kartoza, we can do that by registering `changelog.kartoza.com` for Kartoza organisation. After that the site should only showing the projects that belongs to Kartoza.

### DIFF
--- a/django_project/core/custom_middleware.py
+++ b/django_project/core/custom_middleware.py
@@ -163,5 +163,5 @@ class CheckDomainMiddleware(object):
             else:
                 # for production the domain is hardcoded for consistency
                 return HttpResponseRedirect(
-                    'http://changelog.qgis.org/en/domain-not-found/'
+                    'http://changelog.kartoza.com/en/domain-not-found/'
                 )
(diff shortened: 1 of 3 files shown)


django_project/core/custom_middleware.py
```diff
@@ -163,5 +163,5 @@ def process_request(self, request):
             else:
                 # for production the domain is hardcoded for consistency
                 return HttpResponseRedirect(
-                    'http://changelog.qgis.org/en/domain-not-found/'
+                    'http://changelog.kartoza.com/en/domain-not-found/'
```
timlinux: Could you do a follow up commit that takes this out of settings rather? I don't think it is good to hard code URLs into the code.

django_project/core/tests/test_check_domain_middleware.py
```diff
@@ -54,6 +54,6 @@ def test_invalid_domain_prod(self):
         self.assertEqual(response.status_code, 302)
         self.assertRedirects(
             response,
-            expected_url='http://changelog.qgis.org/en/domain-not-found/',
+            expected_url='http://changelog.kartoza.com/en/domain-not-found/',
```
timlinux: Same as above, Could you do a follow up commit that takes this out of settings rather? I don't think it is good to hard code URLs into the code.

PR #884, opened 2018-04-10
Hide unavailable sections

### LINKED ISSUE #882: Lesson Cosmetics
From this site: http://changelog.inasafe.org/en/inasafe/section/home-assignments/detail/138/

There are some unavailable items in the lessons. We should not show this to the common users or when we print it. So, things to do:

- [ ] Only show available items to the normal user (not staff or admin). We still need to show it to admin/staff so that we can make an edit.
- [ ] When we print it, we should not show any unavailable items (for all users), since we want to print it.

sonlinux:
fix #882 
- [x] Do not show sections to user if it has empty content
- [x] Show action buttons to required users

empty_detail

- [x] Do not show sections to user if it has empty content in PDF print

empty_pdf

- [x] When content is available

normal

### DIFF
--- a/django_project/lesson/templates/worksheet/detail.html
+++ b/django_project/lesson/templates/worksheet/detail.html
@@ -191,6 +191,7 @@
     {# More about #}
 
     <div class="row details-worksheet" style="margin-top:10px;">
+    {% if worksheet.more_about_text or workesheet.more_about_image %}
         <div class="col-lg-1">
           <div class="pull-left" style="background-color:{{ worksheet.section.project.accent_color }};margin-left:-15px;">
             <img src="{% static "img/info.png" %}" style="height:50px;width:50px">
@@ -202,6 +203,8 @@
         <div class="col-lg-3">
         </div>
     </div>
+        {% else %}
+    {% endif %}
     <div class="row details-worksheet" style="margin-top:10px;">
         {% if worksheet.more_about_image %}
             <div class="col-lg-8">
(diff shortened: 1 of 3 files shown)


django_project/lesson/templates/worksheet/detail.html
```diff
@@ -201,57 +203,63 @@ <h3 class="custom-heading" style="margin-left:-40px"><strong>{{ worksheet.more_a
         <div class="col-lg-3">
         </div>
     </div>
+        {% else %}
```
ismailsunni: Do we need this extra `else` if only one case?

django_project/lesson/templates/worksheet/detail.html
```diff
@@ -201,57 +203,63 @@ <h3 class="custom-heading" style="margin-left:-40px"><strong>{{ worksheet.more_a
...
-        </div>
-        {% if worksheet.more_about_image %}
-            <div class="col-lg-4">
-                <a href="#" class="pop-image">
-                    <img id="{{ worksheet.more_about_image.url }}" class="img-responsive img-rounded pull-right"
-                         src="{{ worksheet.more_about_image.url }}"
-                         style="max-width: 100% !important;"
-                         alt="" />
-                </a>
+                {% else %}
```
ismailsunni: Put it in the same indent as the same if statement to make it easier to read

timlinux:
LGTM thanks

PR #887, opened 2018-04-13
Fix worksheet exercise section 

### LINKED ISSUE #882: Lesson Cosmetics

sonlinux:
fix #882 
## Worksheet PDF
pdf_worksheet

## Worksheet Detail
html_worksheet

### DIFF
--- a/django_project/lesson/templates/worksheet/detail.html
+++ b/django_project/lesson/templates/worksheet/detail.html
@@ -110,19 +110,21 @@
     </div>
 
     <div class="row details-worksheet" style="margin-top:10px;">
-        <div class="col-lg-12">
-            <h4><strong>
-                {% trans 'You try:' %}
-            </strong></h4>
-            <p>
-                <strong>
-                    <i>
-                        {% trans 'Goal:' %} {{ worksheet.exercise_goal }}
-                    </i>
-                </strong>
-            </p>
-            <p>{{ worksheet.exercise_task | base_markdown }}</p>
-        </div>
+       {% if worksheet.exercise_goal or worksheet.exercise_task %}
+            <div class="col-lg-12">
+                <h4><strong>
+                    {% trans 'You try:' %}
+                </strong></h4>
+                <p>
+                    <strong>
+                        <i>
+                            {% trans 'Goal:' %} {{ worksheet.exercise_goal }}
+                        </i>
+                    </strong>
+                </p>
+                <p>{{ worksheet.exercise_task | base_markdown }}</p>
+            </div>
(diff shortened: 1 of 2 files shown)


django_project/lesson/templates/worksheet/detail.html
```diff
@@ -110,19 +110,21 @@ <h3 class="custom-heading" style="margin-left:-40px"><strong>{{ worksheet.title
...
+                    {% trans 'You try:' %}
+                </strong></h4>
+                <p>
+                    <strong>
+                        <i>
+                            {% trans 'Goal:' %} {{ worksheet.exercise_goal }}
+                        </i>
+                    </strong>
+                </p>
+                <p>{{ worksheet.exercise_task | base_markdown }}</p>
```
ismailsunni: If a worksheet has only `worksheet.exercise_goal`, but no `worksheet.exercise_task`, it will still show empty line also.
Maybe put also if statement here. And also in the goal part.

django_project/lesson/templates/worksheet/print.html
```diff
@@ -128,11 +128,13 @@ <h2>{{ worksheet.title }}</h2>
...
     {% endif %}
-    <h3>{% trans 'You try:' %}</h3>
-    <p class="goal">
-        {% trans 'Goal:' %} {{ worksheet.exercise_goal }}
-    </p>
-    <p>{{ worksheet.exercise_task | base_markdown }}</p>
+    {% if worksheet.exercise_goal or worksheet.exercise_task %}
+        <h3>{% trans 'You try:' %}</h3>
+        <p class="goal">
+            {% trans 'Goal:' %} {{ worksheet.exercise_goal }}
```
ismailsunni: Same here.

django_project/lesson/templates/worksheet/detail.html
sonlinux: Alright

django_project/lesson/templates/worksheet/print.html
sonlinux: Alright updating

sonlinux:
Updated @ismailsunni 
Thanks

PR #893, opened 2018-04-20
fix sponsor form to check unique fields

ann26:
This PR fixes http://sentry.kartoza.com/sentry/projecta/issues/25/

It fixed the form to check the unique together relations `project` and `sponsor`

gif: when trying to add a new sponsor with the same name within a project
peek 2018-04-20 16-22


timlinux:
Good catch thanks! LGTM!

PR #894, opened 2018-04-20
fix unique together

ann26:
This PR fixes this http://sentry.kartoza.com/sentry/projecta/issues/123/ and http://sentry.kartoza.com/sentry/projecta/issues/125/

It fixes the unique together, that has caused there were duplicates in certificate. 
We also need to delete one of this certificate in database `QGIS-125` or `QGIS-126`. They are duplicates. They have the same attendee and course. I checked it on the course details page the attendee is no longer there. Maybe they have deleted the course attendee but the certificate remains.


ann26:
ok @timlinux , I've deleted `QGIS-125` from both the production and staging database.

PR #906, opened 2018-05-02
Fix version permissions

### LINKED ISSUE #905: Only staff members see the 'create new issue' button
See https://lists.osgeo.org/pipermail/qgis-developer/2018-May/053161.html

When you are a staff member and go to: 
http://changelog.qgis.org/en/qgis/version/3.0.0/
You are able to create a new entry.

But non staff users do not see this button:

selection_413

sonlinux:
fix #905

- [x] Non staff users can create new version

not staff

### DIFF
--- a/django_project/changes/templates/version/detail.html
+++ b/django_project/changes/templates/version/detail.html
@@ -56,13 +56,11 @@
                 {% endif %}
             </div>
             <div class="pull-right">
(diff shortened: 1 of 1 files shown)


timlinux:
Thanks @sonlinux - please don't forget to get your changes peer reviewed before merging!

PR #931, opened 2018-05-25
Add a feed to retrieve the rss feed for past/former sponsors

rduivenvoorde:
@timlinux @ann26 is this ok? To retrieve the past sponsors?

### DIFF
--- a/django_project/changes/feeds/sponsor.py
+++ b/django_project/changes/feeds/sponsor.py
@@ -134,6 +134,28 @@ class RssSponsorFeed(Feed):
         return {'image_url': item.sponsor.logo.url}
 
 
+class RssPastSponsorFeed(RssSponsorFeed):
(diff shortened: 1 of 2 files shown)


timlinux:
Thanks for this @rduivenvoorde  - what about a link in the gui? Happy without it?

timlinux:
@ann26 can you add the past sponsors button and also a test so that our coverage is not reduced? Please make it a meaningful test that inserts some current and expired sponsors and tests we get back the correct set from each feed. Than merge.

rduivenvoorde:
@timlinux I'm ok without a button. I think I'm the only one to use it anyway  :-)

rduivenvoorde:
@ann26 if you give me some examples/pointers I'm happy to add some tests too.

ann26:
Hi @timlinux and @rduivenvoorde I added the button and tests in #940

PR #938, opened 2018-05-26
Add atom url (used by QGIS project), plus

rduivenvoorde:
Some textual fixes (plural 'sponsors')
Add 'past' in feed descriptions for past-sponsors
Changed from '_' to '-' in sponsor url's

@timlinux I forgot that we actually use atom feeds for all feeds in the project... And tested/created only the rss one.
And added some smaller fixes.

### DIFF
--- a/django_project/changes/feeds/sponsor.py
+++ b/django_project/changes/feeds/sponsor.py
@@ -47,7 +47,7 @@ class RssSponsorFeed(Feed):
          :returns: Title of the RSS Feed.
          :rtype: str
          """
-        return 'RSS Sponsor of %s Project' % obj.name
+        return 'RSS Sponsors of %s Project' % obj.name
 
     def description(self, obj):
         """Return a description for the RSS.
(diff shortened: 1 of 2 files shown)


rduivenvoorde:
I still do have some issues with rightfully (responsive) styling of the html in the qgis website.

I know we can 'fix' it, but still. For example: for the past sponsors I would like other class names then current sponsors, I can handle that with more advanced css queries too, but my feeling now is that (at least for sponsors) I maybe have to go for the json route. BUT then json with only key/value(data) pairs, no html in it at all (so probably leave the Feed-stuff and create a json ourselves)....

One other thing:
selection_480

I think we should either only show the sponsor_level (as that also shows the amount_sponsored) OR remove that from the sponsor string.

In the JSON I would like to have separate items for start_date and end_date, so I can so sorting/naming myself. Another advantage of using json is that current strings like: 'Sponsorship leve' and 'Sponsorship period' are then in javascript == translatable in normal website translations.

What do you think?

timlinux:
> I think we should either only show the sponsor_level (as that also shows the amount_sponsored) OR remove that from the sponsor string.

@rduivenvoorde @andreasneumann would like us to support sponsorship thresholds so that a sponsor can give e.g. 600 and recieve bronze status, so we should probably keep that.

> In the JSON I would like to have separate items for start_date and end_date, so I can so sorting/naming myself. Another advantage of using json is that current strings like: 'Sponsorship leve' and 'Sponsorship period' are then in javascript == translatable in normal website translations.

Yes if you make a ticket I can get @ann26 to do that...

timlinux:
Merged - I will deploy tomorrow. Thanks!

PR #940, opened 2018-05-30
add button for past sponsor rss and add test

### LINKED ISSUE #936: We need tweaks for past sponsors
See #931

Can you add the past sponsors button and also a test so that our coverage is not reduced? Please make it a meaningful test that inserts some current and expired sponsors and tests we get back the correct set from each feed. Than merge.

Also please make a view like the current sponsors view which shows all the historidcal sponsors for a project,

ann26:
follow up for #931 and fix #936 

- [x] Added past sponsor rss button
- [x] Added test

The past sponsor button:
screenshot from 2018-05-30 11-37-20

This button is slightly different than the other rss buttons:
peek 2018-05-30 11-38

### DIFF
--- a/django_project/changes/templates/sponsor/list.html
+++ b/django_project/changes/templates/sponsor/list.html
@@ -67,6 +67,11 @@
                        data-title="JSON Feed for Sponsors">
                         <i class="fa fa-rss-square"></i>
                     </a>
+                    <a class="btn btn-default btn-mini tooltip-toggle rss-icon"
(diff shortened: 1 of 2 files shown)


ann26:
Yes @ismailsunni :p 
any better ideas?

timlinux:
Its ok for now - better to have three symbols than to have hidden functionality.

timlinux:
Thanks @ann26 cc @rduivenvoorde

PR #951, opened 2018-06-06
Remove pending version state.

### LINKED ISSUE #806: Remove Pending State for Version, Category, and Entry
It turns out the approach to have pending state doesn't work properly. After discussion, @timlinux suggests removing the pending state for:
- [x] Version
- [x] Category
- [x] Entry

It will have these consequents:
- [x]  All normal user can not edit or add anything
- [x] Edit will happen directly after click save (no more pending state)
- [x] Only staff, superuser, project owner and version manager can make edits

ismailsunni:
Fix #806


django_project/changes/tests/test_models.py
```diff
@@ -193,8 +192,8 @@ def test_formatted_release_date(self):
         """Tests we can get, set and present the release date nicely."""
         model = VersionF.create(
             description=u'New description',
-            release_date=datetime(2016, 06, 06),
-            approved=True)
+            release_date=datetime(2016, 06, 06)
```
ismailsunni: __[Flake8]__

[E999] SyntaxError: invalid token

ann26:
LGTM, but the codecov is decreasing

ismailsunni:
> LGTM, but the codecov is decreasing

The codecov is decreasing since there was not unit test previously on the file that I changed.

timlinux:
LGTM

django_project/changes/feeds/version.py
```diff
@@ -78,7 +78,7 @@ def items(self, obj):
         :param obj: A project
         :type obj: Project
 
-        :returns: List of approved version of a project
+        :returns: List of all version of a project
         :rtype: list
         """
         return Version.objects.filter(project=obj, approved=True).order_by(
```
Gustry: approved?

ismailsunni: Thanks, @Gustry !!

PR #953, opened 2018-06-06
Fix logic in version list template.

ismailsunni:
And make the code a little bit better

screen shot 2018-06-07 at 05 43 36


ann26:
LGTM

PR #967, opened 2018-06-29
Fix pending approval permission

### LINKED ISSUE #950: Pending menu is visible for anonymous users
I'm not connected:
screenshot from 2018-06-04 11-53-52

I don't think that this menu should be visible for anonymous. All items are protected.

sonlinux:
fix #950 

pending_approval


timlinux:
LGTM thanks @sonlinux

PR #971, opened 2018-07-09
Resolve incorrect URL

### LINKED ISSUE #949: http:// must be added manually
If in founder URL of an changelog entry, the URL is 'www.something.com', the link is wrong. The user must add 'https://' to make it work properly. I think projecta should handle this ? It might be the same issue in developer URL.

sonlinux:
fix #949 
Automatically Resolve URLs to **http://www.somlink.com/** if link is **www.somelink.com** and leave 'as is' if it is already prefixed with a scheme.

resolve_url


timlinux:
LGTM thanks @sonlinux

PR #987, opened 2018-10-24
fix certification about page

ann26:
hi @timlinux ,
Alison might have accidentally changed the template.


timlinux:
Great, thank you so much @ann26 !

PR #988, opened 2018-10-24
make answer publicly available

ann26:
hi @timlinux with this PR I made the answers publicly available.

Here's gif when user is not logged in:
peek 2018-10-24 15-33


timlinux:
Thank you so much @ann26 and FYI @anitagraser - I will deploy this just now...

PR #989, opened 2018-10-24
add sign up button

### LINKED ISSUE #947: We need a clear call to action on the certification page.
I am getting emails from people asking how to sign up for certification. Can you put a nice big button on the certification page saying 'sign up'

I think they don't recognise the start metaphor for a new org and also they need to see the call to action even if they are not logged in.

http://changelog.qgis.org/en/qgis/certifyingorganisation/list/

ann26:
fix #947 

I added the sign up button and about certification icon. These two icons are available for public.
When non logged in user click on sign up button, it will redirect them to login page then after login they'll see the create a new certifying organisation form.

Public:
peek 2018-10-24 16-01

Logged in user:
screenshot from 2018-10-24 16-02-59

### DIFF
--- a/django_project/certification/templates/certifying_organisation/list.html
+++ b/django_project/certification/templates/certifying_organisation/list.html
@@ -22,29 +22,39 @@
         <h1 class="text-muted">
             {% if unapproved %}Unapproved {% endif %}
             Certifying Organisations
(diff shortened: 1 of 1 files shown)


timlinux:
LGTM thanks

PR #990, opened 2018-10-24
show only current sponsor and add future sponsor list

### LINKED ISSUE #972: We should not show 'future' sponsors
See https://github.com/qgis/QGIS-Website/issues/550#issuecomment-405466754

Sometimes there are sponsor records in future, we should create a  query to NOT have those in results

anitanh:
fix #972 

- [x] Sponsor list page only list current sponsor and past sponsor
- [x] Added future sponsor list (can only be seen by staff, owners or managers)

peek 2018-10-24 17-03


timlinux:
@andreasneumann @rduivenvoorde Can you guys confirm this behaviour is now correct?

timlinux:
Thanks for the PR @ann26 !

timlinux:
@ann26 can you please review which parts of this PR are still valid  (be careful not to break the sponsor feeds!) and maybe just merge the future view if we have already the present filtering all in place.

PR #995, opened 2018-11-09
add lock version feature

### LINKED ISSUE #973: [FEATUREQUEST] we should be able to 'lock' editing of QGIS-version changelog after release
Currently I'm not aware of disabling the editing of items in the changelog list.

But as an example, now we have exported/released the visual changelog for QGIS 3.2 I think it is not nessecary anymore that people update the changelog records for 3.2 ?
The edits will never show up in the website anymore, so it is a wast of time to edit those.

So: feature request: an admin should be able to 'lock editing' of a version, so people cannot edit anymore, and if possible there is a small message on top: 'editing is disabled for this version' or so?

anitanh:
fix #973 

- [x] Add lock and unlock function that is only available for staff user

Lock version:
peek 2018-11-09 10-06

Unlock version:
peek 2018-11-09 10-06 2

View for non staff user but not public user (e.g changelog managers):
Locked version:
screenshot from 2018-11-09 10-11-43

Unlocked version:
screenshot from 2018-11-09 09-29-03


timlinux:
Lovely thanks!


What do you think - should we get rid of these codecov tests?

ann26:
hi @timlinux I think so, sometimes getting it passed is a pain :sweat_smile:

django_project/changes/views/version.py
```diff
@@ -22,7 +22,7 @@
     DetailView,
     UpdateView,
 )
-from django.http import HttpResponse
+from django.http import HttpResponse, HttpResponseForbidden
```
ismailsunni: __[Flake8]__

[F401] 'django.http.HttpResponseForbidden' imported but unused

timlinux:
@ann26 please confirm that project managers are able to lock and unlock versions and then please sync, test and merge.

PR #1002, opened 2019-02-05
Set version for WeasyPrint.

ismailsunni:
It's needed to avoid installing the latest WeasyPrint that only supports Python 3

### DIFF
--- a/README-dev.md
+++ b/README-dev.md
@@ -107,7 +107,8 @@ Now set these options:
 
 * **Name:** Django Server
 * **Host:** 0.0.0.0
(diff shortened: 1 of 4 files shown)


ismailsunni:
The failed in travis is not because of this PR (most probably), maybe @ann26 knows about it :)

ann26:
hi @ismailsunni set it as `WeasyPrint>=0.42,<0.43` that'll do the trick :)

ann26:
@ismailsunni ah sorry, you've fixed that, (I've just checked travis). I think it's because the test sponsorship end date is outdated (see https://github.com/kartoza/prj.app/blob/44090a910ea92d5117fab33ddd4b3553fa6a82b9/django_project/changes/tests/test_rss_view.py#L52)

PR #1008, opened 2019-03-07
Update Python Package Version

ismailsunni:
Fixing the error in local and travis

```
regex._regex_core.error: bad character range at position 294 (line 7, column 55)
```


ismailsunni:
Already included in #1010

PR #1033, opened 2019-09-04
add status and remarks for certification review process

ann26:
This PR:
- [x] Add Status model, where admin can add/edit/remove status for reviewing certifying organisation. 
- [x] In pending certifying organisation view, we can edit the status of a certifying organisation and give remarks.

Editing status and remarks
Peek 2019-09-04 19-32

We can also approve the organisation from this form as long as we have approved status on the list
Peek 2019-09-04 19-32 2

Status can be added/edited/removed from admin. 
The order number is the order that will be displayed in the form.
The Status has foreign key with project and it's name is unique within the project (name and project are unique together).
Peek 2019-09-04 19-33

After updating the Status model in admin, we can see in front end.
Peek 2019-09-04 19-40

### DIFF
--- a/deployment/docker/REQUIREMENTS.txt
+++ b/deployment/docker/REQUIREMENTS.txt
@@ -36,3 +36,4 @@ django-modeltranslation==0.12.2
 django-colorfield
 WeasyPrint==0.42.3
 djangorestframework==3.6.4
+django-simple-history==1.9.1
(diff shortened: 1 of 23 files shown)


timlinux:
Ah that looks great @ann26 ! Does that tick all your boxes @geomenke?

geomenke:
This looks good. The one additional thing that would be helpful is the date contacted

timlinux:
> This looks good. The one additional thing that would be helpful is the date contacted

Thanks yeah I asked Anita to add that too...

ann26:
hi @timlinux and @geomenke 
I added django-simple-history to track changes, as shown below:
Peek 2019-09-06 14-23

Using this, you can also revert the certifying organisation object to the version you like to come back to, like this:
Peek 2019-09-06 14-25

What do you think?

timlinux:
Very nice @ann26 and simple-history is great! LGTM to merge.

PR #1037, opened 2019-09-06
only enable issue button when credits are sufficient

### LINKED ISSUE #1031: Negative credits
image
Should this be possible?

ann26:
fix #1031 

This PR:
- [x] Change the logic on how to issue certificates based on available credits. 
Previously, we can issue certificates as many as we need regardless the amount of the available credits but we cannot print them when the credits are not sufficient, thus we can have negative credits. In the previous logic, once we have sufficient credits, we can pay the issued certificates but not printed using the credits.
- [x] Add new logic: when the credits are not sufficient, the button to issue certificate will become the button to inform user that their credits are not enough and need to inform the administrator.
Peek 2019-09-06 17-46
- [x] Add custom validation to check available credits upon issuing the certificate.
Peek 2019-09-06 18-27

### DIFF
--- a/django_project/certification/forms.py
+++ b/django_project/certification/forms.py
@@ -6,6 +6,7 @@ from django.contrib.auth.models import User
 from django.contrib.gis import forms as geoforms
 from django.contrib.gis import gdal
 from django.contrib.gis.forms.widgets import BaseGeometryWidget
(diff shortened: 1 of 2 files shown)


timlinux:
Please fix and merge when ready.

PR #1067, opened 2019-10-30
remove disqus from pages

ann26:
Removing disqus because of unwanted ads.


timlinux:
Thanks! LGTM! Please deploy when ready .... CC @mbernasocchi

ann26:
@timlinux @mbernasocchi deployed on staging and live

PR #1071, opened 2019-11-06
improve certificate layout

### LINKED ISSUE #1070: Improve certificate
The font should be the same and the spaces should be even

ann26:
fix #1070 

Screenshot from 2019-11-06 17-33-17


timlinux:
Much better, thanks @ann26 !

PR #1087, opened 2019-11-24
always show button

### LINKED ISSUE #1083: Buttons to go back should be visible
In the lesson app:
http://changelog.qgis.org/en/qgis/section/data-providers-21/detail/79/?q=2.1

I know why these buttons are hidden by default. But this a UX bug, as people have no idea that they need to put their cursor in this area to display the back button. And honestly, for the end user, it's not convenient/nice for the display when it's moving up/down.

A flag is needed to display these buttons when we are in the webbrowser or when we are printing a PDF.

ann26:
fix #1083 

Screenshot from 2019-11-24 17-03-36

### DIFF
--- a/django_project/lesson/templates/worksheet/detail.html
+++ b/django_project/lesson/templates/worksheet/detail.html
@@ -26,14 +26,14 @@
     <div class="row page-header" style="border-bottom: 8px solid {{ worksheet.section.project.accent_color }};">
         <div class="col-lg-12">
             <div class="pull-right">
-                <div class="btn-group worksheet-action" style="display:none;">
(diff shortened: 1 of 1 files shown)


Gustry:
@ann26 These buttons were hidden by default because we don't want them in the PDF. Is-it the case with this fix?

ann26:
@Gustry the pdf has different html template here https://github.com/kartoza/prj.app/blob/develop/django_project/lesson/templates/worksheet/print.html

timlinux:
Thanks @Gustry and @ann26, yeah it is probably better to show the buttons if they dont affect the PDF output.

PR #1113, opened 2020-01-27
Fix zip download

dimasciput:


dimasciput:
Some fixes due to python 3 upgrade

PR #1133, opened 2020-02-11
add function to import PR into entries

### LINKED ISSUE #1128: We need to be able to stub in changelog entries from PR's
When a PR is made in the project repo, under specific conditions, we should make a changelog entry. Those conditions include:

1. The changelog Version model does not have an enabled 'watch PRs' Attribute. The Watch PR's attribute should be managed that only one version at a time can have this flag active. So creating a new version and ticking this flag should disable watches against the previous version. Any incoming notification hooks should be applied against the currently active Version. 

image

image

2. The PR contains one or more matching tags.  See images below for some tag example. The tag match list should be defined by project managers in the project properties dialog, as should the github URL:
image

3. The title of the PR should be used for the changelog entry title

4. The description of the PR should be used for the changelog entry body (including all markdown)

5. An upstream_pr property should be added to the Entry model that links to the upstream PR which should be rendered as a hyperlink of the changelog view.

6. The author of the Entry should be populated by doing a lookup on their GH profile to fetch their full name. By default it can be a hyperlink to their GH user page.

There has been previous similar work carried out for the QGIS documentation team, so your first stop should be: 

https://github.com/qgis/QGIS/blob/master/.github/workflows/pr_to_doc_issue.yml

Contact @3nids for more info on above if needed

ann26:
fix #1128 

This PR:
- [x] Added function to fetched PRs from github
- [x] Added modal in version detail view to import PRs from github
- [x] Added form validation

Note:
- On deployment we need github token so that we can have request to github API up until 5000 request per hour, without this token we only have 60 requests per hour 
- for rate limiting, see: https://developer.github.com/v3/#rate-limiting.
- to set up github token see: https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line#creating-a-token

Peek 2020-02-11 12-06

## How to use
1. Click on the github icon on the right side of the page, it will open a pop up to import the PRs.
2. The repository url is filled automatically but should you like to import the PR from another repository you can fill it here too (<b>required</b>).
3. Wait until the label options is rendered.
4. Select the label of which you want to import the PR, if you select more than one label it will only fetch the PR containing all of the selected labels.
5. Select the state of the PR, it can be open or closed or both, if you don't select any it will fetch all PR regardless open or closed.
6. Select a category where the new entries will be associated to (**required**).
7. Click import, it will begin fetching your PRs and rendering them into entries.

### DIFF
--- a/django_project/certification/views/certificate.py
+++ b/django_project/certification/views/certificate.py
@@ -419,7 +419,7 @@ def generate_pdf(
         margin_left, (margin_bottom - 20),
         'You can verify this certificate by visiting '
         'http://{}/en/{}/certificate/{}/.'
-            .format(current_site, project.slug, certificate.certificateID))
+        ''.format(current_site, project.slug, certificate.certificateID))
 
     # Close the PDF object cleanly.
     page.showPage()
@@ -626,7 +626,7 @@ def email_all_attendees(request, **kwargs):
                 '{organisation_slug}/course/'
                 '{course_slug}/print/{pk}/\n\n'
                 'Sincerely,\n{convener_firstname} {convener_lastname}'
-                    .format(**data),
+                ''.format(**data),
                 course.course_convener.user.email,
                 [attendee.email],
                 fail_silently=False,
--- a/django_project/changes/admin.py
+++ b/django_project/changes/admin.py
@@ -55,6 +55,12 @@ class VersionAdmin(reversion.admin.VersionAdmin):
 class EntryAdmin(reversion.admin.VersionAdmin):
     """Entry admin model."""
 
+    list_display = ['pk', 'version_name', 'category', 'title']
+    list_filter = ['category']
+
+    def version_name(self, obj):
+        return obj.version.name
+
     def queryset(self, request):
         """Ensure we use the correct manager.
 
--- a/django_project/changes/models/sponsorship_period.py
+++ b/django_project/changes/models/sponsorship_period.py
@@ -61,7 +61,8 @@ class SponsorshipPeriod(models.Model):
     )
 
     currency = models.CharField(
-        help_text=_('The currency that is used for sustaining membership payment.'),
+        help_text=_('The currency that is used for '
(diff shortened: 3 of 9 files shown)


ann26:
Travis error is unrelated to this PR

dimasciput:
Hi @ann26 , shouldn't this be like an automated process? User doesn't need to do this manually, instead when there is a new PR with specific label, it will create an entry in changelog? 
What do you think @timlinux ?

timlinux:
Hi @ann26  and @dimasciput 

I'm OK with it being a manual process for now but there are a few more things maybe I missed to mention in the original ticket:

We need to be able to specify a time box or at least a starting point so that we don't fetch PR's from before the current release cycle.

It's not clear from your GIF but do you bring image references through?

Do we have logic in place to prevent duplicate entries?

Ideally in the future ,yes, I would like the process to be automated, with the options you provide in the dialog moved into the project properties page so that any new incomiing PR's get added to the latest current changelog. We need a flag to mark a version 'open' or 'closed' to indicate whether new entries may be added to the latest changelog. I would keep the manual tool around too.

ann26:
hi @timlinux 

> It's not clear from your GIF but do you bring image references through?

by image references do you mean the image in the PR? yes it will be in the entries too.


>  Do we have logic in place to prevent duplicate entries?

No, but if we want to create one, how do we determine that it is a duplicate? If it is the title, I think it is possible to have the same title and yet different PR e.g. update on feature etc

> We need to be able to specify a time box or at least a starting point so that we don't fetch PR's from before the current release cycle.

I will add the time box

timlinux:
> No, but if we want to create one, how do we determine that it is a duplicate? If it is the title, I think it is possible to have the same title and yet different PR e.g. update on feature etc

Simply having the same PR url - can you add a PR url field to then entry model for this purpose:

image

ann26:
@timlinux ah okay, I will do, I will create a separate PR with the time constraints and duplicate handler later

PR #1141, opened 2020-02-17
change font to support turkish characters

### LINKED ISSUE #1043: Generated certificate special characters issues for course convener
The generated certificates support special characters except for the course convener (probably other font).
See:
[wrong_characters.pdf](https://github.com/kartoza/prj.app/files/3640982/wrong_characters.pdf)

I manually edited the certificates to bypass this.
Maybe changing the font of the course convener would be a better solution?

ann26:
fix #1043 

"Times-Italic" font does not support turkish characters, so I changed it to use the regular one, "Noto-Regular" is it okay @timlinux ?

before:
Screenshot from 2020-02-17 21-31-30

after:
Screenshot from 2020-02-17 21-31-39

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,4 +181,7 @@ Thank you to the individual contributors who have helped to build projecta:
 * Ismail Sunni : imajimatika@gmail.com
 * Richard Duivenvoorde : richard@duif.net
 * Rischan Mafrur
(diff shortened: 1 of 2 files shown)


timlinux:
+1 to use non italic thanks

PR #1143, opened 2020-02-19
fix preview certificate

### LINKED ISSUE #1142: preview certificate is not working
from a user:

I'll be doing my first QGIS Training in the Philippines next month and was trying out using certificate templates in Projecta to create more personalized certificates for my attendees.

Previously, I was able to preview custom certificate templates by following the instructions here: https://changelog.qgis.org/en/qgis/about/.

But now, whenever I try to use a custom template, I get an error 500 like the one below:
image

I've tried this with different images -- transparent, all-white, etc. -- but the result is the same. Previewing a certificate without a template works though.

ann26:
fix #1142

### DIFF
--- a/django_project/certification/views/certificate.py
+++ b/django_project/certification/views/certificate.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 
+import base64
 import datetime
 from io import StringIO, BytesIO
 import os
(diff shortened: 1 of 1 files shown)


timlinux:
Thanks - let me know when it is deployed so I can write to Ben.

Regards

Tim

ann26:
@timlinux deployed to production

PR #1161, opened 2020-03-05
add user profile page

### LINKED ISSUE #1148: Only GH username is shown for Course Convener (not full name) so the name does not show in certificates
Screenshot_20200223_153635

The GH username of this person is "andremano", he has a public full name on GH, but the certificate platform does not show his full name, so when generating certificates the name does not show.

ann26:
fix #1148 

@timlinux what do you think?
Peek 2020-03-05 17-52
Note: only the user itself can open the edit page for his profile.


timlinux:
Perfect thanks @ann26 !

Issue #1216, opened 2020-10-06
After updating the splash in 3.16 it is not visible anymore


rduivenvoorde:
I thought to update the splash with the upcoming one from Google drive of Anita,
but after editing (the version) it is not visible anymore in the site:
https://changelog.qgis.org/en/qgis/version/3.16/
https://changelog.qgis.org/media/images/projects/dbc0cef663f2f652e0961ac2ed168eb45ed2e182.png gives an 'forbidden'..

zacharlie:
I have had this issue with media on the changelog. Sometimes (although not always) uploading the image again resolves it... Still an issue that needs to be resolved but the "turn it on and off again" routine seems to be a possible workaround.

rduivenvoorde:
Just tried to upload it again, it is current splash:
https://drive.google.com/open?id=1GmSpwppbX9dIjYrf3n_ZelC97Bv-lRYU

But it still does not show up:

![Screenshot-20201006120201-1348x445](https://user-images.githubusercontent.com/731673/95187708-c2a0b680-07cb-11eb-8b99-a8ddd94d8cb6.png)

url is now: https://changelog.qgis.org/media/images/projects/dbc0cef663f2f652e0961ac2ed168eb45ed2e182.png

zacharlie:
I see. I tried as well and when it redirected to the versions list I see the thumbnail is available... bu I also get a permissions error (403) when trying to access the actual file. Might be a webserver config issue... Or a user permissions problem...

![image](https://user-images.githubusercontent.com/64078329/95191070-83289900-07d0-11eb-8525-9e63f450bbdf.png)

timlinux:
So it is a bug in that if you change the splash or version text, the version gets unpublished. @sumandari can you make sure the publish status remains intact after doing any edits?

rduivenvoorde:
@timlinux what do you mean 'get unpublished'? I do not think 3.16 got unpublished, did it?

https://changelog.qgis.org/en/qgis/version/3.16/

Just the permissions of the image file seem wrong?
As @zacharlie says (and shows in the screenie above), 3.16 is in the Versions list, AND shows the right splash (there!)...
I think @zacharlie ment to say that he get's a permission error when he tests the image url

timlinux:
Ok I fixed the image. @sumandari the bug is actually if you save the version description in the frontend editor, it drops the release date. If you save it in django admin, it preserves the date.

So: 

[...]

PR #1222, opened 2020-10-12
assigned value to release_date attribute on save method

sumandari:
@timlinux 

This PR is referring to issue #1216 and https://github.com/kartoza/prj.app/issues/1216#issuecomment-704254144

**What does this PR do:**
This PR assigns the value of `release_date` attribute in modelForm when the instance invoking `save()` method, so that if it edits the version in the frontend editor, it will preserve the date.

_**gif image:**_
projecta_1216


timlinux:
LGTM (Looks Good To Me)

PR #1229, opened 2020-11-08
Update changelog_github.py

dimasciput:
Change funded_by_url to funder_url


Gustry:
Sorry for that. Thanks for the fix.

Issue #1236, opened 2020-12-21
evaluate and change downloads behaviour


zacharlie:
Some strange behaviour has been observed when interacting with platform media and downloads, however these may have been infrastructure related or pertain to legacy content. Regardless, the download functionality should be evaluated carefully by all parties concerned to identify any possible bugs or UX enhancements.

Currently, when downloads do work as expected it seems that course contents are evaluated as string literals and contain special characters, e.g. ```r'FooBar.zip'```, whilst pdf downloads are simply output as the initial file content. All downloads should be output with a meaningful title, and a file extension relevant to the content MIME type.

sumandari:
Hi @zacharlie, thanks for the detail. It help me found the naming issue in the code. 
https://github.com/kartoza/prj.app/blob/cff9bdb21be92ccfa536d15638a2387435fc8fd4/django_project/lesson/views/worksheet.py#L139
I think decode the string (file_name) will remove `prefix b'` in the filename. I tried on my local machine:
![changelog_1236_downloaded_filename](https://user-images.githubusercontent.com/40058076/102866367-19846680-4472-11eb-9f1e-6c91d30d33b1.gif)

Could you please confirm (and provide example case) if you have experienced with file extension that is not relevant to the content MIME type, so that I can fix it?

zacharlie:
As an example, when downloading a pdf, I would expect to be able to save the file and have the system prompt me to save it as *1.1. QGIS User Interface Tour.pdf*, but instead it requests *1.1. b'Introduction to QGIS-QGIS User Interface Tour'. b'Introduction to QGIS-QGIS User Interface Tour'* of an unknown file type

![image](https://user-images.githubusercontent.com/64078329/102869901-f2fc0680-4444-11eb-90be-af1cc40018e0.png)

NyakudyaA:
Another issue related https://github.com/kartoza/prj.app/issues/914

sumandari:
Thank you @zacharlie @NyakudyaA , I hope extracted zip from my gif animation will fix the issue. I will submit the PR so you can review it. 
![changelog_1236_pic0](https://user-images.githubusercontent.com/40058076/102870933-af22f480-4478-11eb-88fb-ee7198ac8a06.png)

```
suman@NUC10:~/Downloads/QGIS-worksheet module 1.1 to 1.2 (1)$ tree .
.
└── 1. Introduction to QGIS
    ├── 1.1. QGIS User Interface Tour.pdf
[...]

Issue #914, opened 2018-05-10
data download files should have better names


gubuntu:
like <worksheet title>_data.zip

NyakudyaA:
Tested with file http://geonode.kartoza.com/layers/geonode:flood_laplata_classified_2013_3 and the data downloads had nice names.
![downloads](https://user-images.githubusercontent.com/2510900/40232444-1264f4c4-5a9f-11e8-8df3-917714374562.png)

For the Geotiff it would be nice if all the layers downloaded had an extension added to them ie `flood_laplata_classified_2013_3.tif`

gubuntu:
### lesson data file download file names badly constructed

The file download at https://changelog.qgis.org/en/qgis/section/symbology-22/detail/81/?q=3.1 provides a valid zip file but it has a badly constructed name, which has no .zip extension

```b'Symbology-Styling basics'```

Fix this and the download links for other worksheets

PR #1244, opened 2020-12-22
fix downloaded filename

sumandari:
This PR refers to #1236 and #914. It does:
- fix downloaded zip filename
- fix downloaded pdf filename

914_naming


zacharlie (approved):
LGTM

Issue #1246, opened 2020-12-22
Text in PDF are not rendered properly


NyakudyaA:
On-site

![site-a](https://user-images.githubusercontent.com/2510900/102887432-9fe37d00-445f-11eb-8a5b-00f82faaa774.png)

Downloaded PDF

![down-pdf](https://user-images.githubusercontent.com/2510900/102887465-b4277a00-445f-11eb-903a-f9cfac986d4a.png)

zacharlie:
I have been going through and manually replacing these characters in the content and using the ```'``` character on my windows system doesn't seem to have these errors. Manual fix is still a work around though so I definitely think this should be prioritised.

NyakudyaA:
I think some of this is because the worksheets initially were written in libre office and someone had to copy-paste the content from the libre office docs to the web.

sumandari:
@NyakudyaA the fix has been deployed. Please check out the example case:
https://changelog.kartoza.com/en/geonode/section/geonode-28/print/111/?q=1.2.1
Can we close the issue now?

PR #1251, opened 2020-12-23
Set encoding utf-8 in lesson pdf output

sumandari:
@NyakudyaA @zacharlie 
This PR refers to #1246. It specifies the character encoding with value "UTF-8"
1246_set_encoding


NyakudyaA:
LGTM

Issue #1233, opened 2020-12-21
Add a "funded by" element to lessons


zacharlie:
Add a "funded by" element, the same as which is available in the changelog entries, but for lesson modules.

sumandari:
Hi @zacharlie, the fix #1253 has been deployed.

zacharlie:
I can't seem to find how this value is modified.

sumandari:
This element cannot be modified by author yet. Would you please explain should it be an input field so that the author can fill it in or it's a dropdown shows list of funder from the database?

zacharlie:
I think an input field embedded directly on the edit page for each module is probably the best approach...

sumandari:
The fix #1272 has been deployed on staging server.

PR #1253, opened 2020-12-24
Add a "funded by" element to lessons

sumandari:
This PR refers to #1233. It does:
- add a query to fetch funded_by and funder_url field in entries filter by last version of project
- display it on detail page (web) and print page (pdf)
- fix youtube.png image, so that it has the appropriate background color

@zacharlie please confirm if the queryset meets your expectation. I was not sure about which funder should be displayed. Thanks! :-)

web view

1233_funded_by_1
---

pdf view

1233_funded_by_2
1233_funded_by


zacharlie (approved):
I think this is fine for now 👍

Issue #1234, opened 2020-12-21
"Add New" for "Further reading" elements


zacharlie:
Currently it seems the only way to create further reading elements is via the django admin interface. This may possibly be a permissions error, but it is required that further reading items are able to be added from the module page

sumandari:
The fix #1254 has been deployed on staging server.

PR #1254, opened 2020-12-24
always enable create further reading

sumandari:
This PR refers to #1234. It does:
- always enable create new further reading 
- reverse the order of further reading item

changelog_1235_lesson_further_reading


timlinux:
LGTM

Issue #1239, opened 2020-12-21
Enhance PDF outputs


zacharlie:
Aside from the issues outlined in #1236 , PDF outputs should be further enhanced for print materials by customising the styling used to generate PDFs

Currently the following specific items need to be addressed:

- List items margins/ left padding needs to be reduced to be in line with paragraph content
- Table rendering issues need to be addressed. Currently tables overlap with associated media elements
- Section icons need to be standardised. Currently unicode symbols are rendered differently between browsers and between different pdf outputs, such as the book symbol from the further reading element.

sumandari:
@zacharlie the fix has been deploy. Can we close the issue or is there another thing should be improved?

PR #1259, opened 2020-12-30
enhance pdf output

sumandari:
This PR refers to #1239. It does:
- change items list padding to be in line with paragraph content 
- fix overlapping media elements in question section
- change book unicode symbols to book image


zacharlie (approved):
Thanks this seems to address all my immediate concerns

PR #1262, opened 2021-01-05
fix 500 after login with email in prod

sumandari:
This PR to fix 500 error temporarily. It forces to redirect to /en/ after user logged in with username/email.

### DIFF
--- a/django_project/core/urls.py
+++ b/django_project/core/urls.py
@@ -6,6 +6,7 @@ from django.contrib.auth import views as auth_views  # noqa
 from django.contrib import admin
 from django.conf import settings
 from django.conf.urls.static import static
+from django.views.generic.base import RedirectView
 from django.http import HttpResponseServerError
 from django.template import loader
 from .views import general_flatpage
@@ -32,7 +33,11 @@ def handler500(request):
     }))
 
 
-urlpatterns = []
+urlpatterns = [
+    # force redirect to /en/
+    # to fix 500 error temporarily  after user logged in on production
+    url(r'^$', RedirectView.as_view(url='/en/')),
+]
 # These patterns work if there is a locale code injected in front of them
 # e.g. /en/reports/
 urlpatterns += i18n_patterns(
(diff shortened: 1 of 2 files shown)


lucernae:
@sumandari I've found some reference to set the default activated language in this page:

https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#explicitly-setting-the-active-language

I propose that the root `'^$'` url uses a custom view function where it activates the current language in the cookie session. If the cookie doesn't have language set yet, then use default language specified in `settings.LANGUAGE_CODE`, which is probably `en`. After activating the language, the view function then returns a redirect to the reversed url of the index page (the one in the `base.urls`)

django_project/core/urls.py
```diff
@@ -32,7 +33,11 @@ def handler500(request):
     }))
 
 
-urlpatterns = []
+urlpatterns = [
+    # force redirect to /en/
+    # to fix 500 error temporarily  after user logged in on production
+    url(r'^$', RedirectView.as_view(url='/en/')),
```
lucernae: Can you try the following:

Replace this with:

```suggestion
    url(r'^$', index_view),
```

Where `index_view` defined as

```python
from django.utils import translation
from django.shortcuts import redirect
from django.urls import reverse

def index_view(request):
    user_language = translation.get_language() or settings.LANGUAGE_CODE or 'en'
    translation.activate(user_language)
    # I looked up the view name in base.urls
    return redirect(reverse('home'))
```

lucernae: It might not work out of the box, but try and modify if possible.

lucernae (commented):
Changes to set default language for the index view

django_project/core/urls.py
sumandari: Thanks @lucernae, I implement your code in the new commit. I gave it a try on my local machine and it worked as expected.

sumandari:
The issue fixed by #1269

sumandari:
reopen the PR, the error still exists

lucernae:
LGTM. Try it in staging with DEBUG False so we can be sure.

PR #1270, opened 2021-01-07
Add action to build docker image for PR

lucernae:
The images will be built with the format `kartoza/projecta-uwsgi:pr-<pr id number>`

Useful if you have multiple PR and multiple image to test.


lucernae:
Sorry for the spam commit. I'm editing from Github web. Will squash on merge later.

So we have a built in github action security feature that doesn't allow fork's PR to access upstream repo's secrets. But it doesn't access requestee's secrets either. So it seems, this just not possible from PR outside of the repo.

timlinux:
Is this done? Can we merge? Please make sure docs are updated too.

lucernae:
Sorry, forgot to follow up.

In the end, I think the use case is different from what I imagined.
We cannot auto build docker image if the PR is coming from fork.
But we can autobuild PR for internal branches (useful if we have feature branches).
Not sure if this is still useful though.

Issue #1240, opened 2020-12-21
Add print controller elements to worksheets


zacharlie:
Related to the issues in #1239 , it would help if print-only controller elements were introduced to worksheet layouts for improving pdf export quality.

Immediate use cases are the inclusion of page break elements, or the definition of media/ content sizes to ensure that screenshots etc. can be resized when generating pdfs.

sumandari:
@zacharlie I added a css class for page break element in #1265. But I have no idea, how to add an user interface to define the image/content size. Should it be a new input field (model field) to set  all image size, or maybe in another way. Could you please give a suggestion?

zacharlie:
I think we should add a definition to each image. For example, on the edit page, under the "browse button" for "Summary image", we could add a **Dimensions (optional): Width [] Height []**. The CSS can then be applied based on an element id (summary-img/exercise-img etc) and will only use these values if they are specified. Perhaps a "clear dimensions" button would help as well.

sumandari:
The fixes #1265 and #1273 has been deployed on staging server.

zacharlie:
I notice the page break element is only usable within content elements. Whilst this is somewhat useful for pushing lists etc onto the new page, it is very limited in function.

It is presenting a problem as I cannot place a break before the start of an element (the header will be on one page, but the content will be forced onto the next), or at the end (the image/ media that follows will be pushed onto the next page)

Can we try think of a way to resolve these issues?

sumandari:
> he header will be on one page, but the content will be forced onto the next

@zacharlie could you please give a screen capture/ illustration image of the example case you mentioned above. Thank you :sunflower:

sumandari:
![image](https://user-images.githubusercontent.com/40058076/105679783-e2e6b380-5f29-11eb-980b-2385224bf773.png)

Do you mean, we need to enable the page break between the image and `Check your knowledge`, also between exercise question and `Further reading`?

zacharlie:
Yes precisely, apologies for delay

sumandari:
This fix has been released on v2.3.0 and deployed on prod. I will close this for  now

PR #1273, opened 2021-01-11
set dynamic css for image in pdf output

sumandari:
This PR refers to #1240 to set dimension of image in PDF output. It does:
- add summary_image_dimension, exercise_image_dimension, more_about_image_dimension fields on lesson model.
- set css value of images in PDF output 

image


timlinux:
LGTM

Issue #1241, opened 2020-12-21
Edit worksheet table headers


zacharlie:
Currently it seems that worksheet tables have static header values, which should be able to be edited

sumandari:
The fixed has been deployed, I will close this ticket now

Issue #847, opened 2018-03-12
requirements table questions / tweaks


gubuntu:
1. 'expectation' doesn't seem to be a generally useful field name. How about 'value'
1. What happens with the Notes field in the requirements table? It doesn't appear in the worksheet but space seems to be allocated to a column for it.

NyakudyaA:
@sumandari I suggest you tackle issue 2 above because with 1 you would need to discuss it with Tim

Maybe we could have the Notes as a tooltip but we would need to let users know that if they hover over the requirements they will see a tooltip if available.

Do we need to print the notes into the PDF?
ping @zacharlie

zacharlie:
It's hard to say whether the notes value would be useful in the printed materials because that would probably depend on the content, but yes if it is intended as a tooltip describing the column values then I would recommend using the notes as a description that subtext in the printed materials. I'm attaching a screenshot outlining my intention below.

![image](https://user-images.githubusercontent.com/64078329/102881831-1bd8c780-4456-11eb-9183-ffc1cf44f624.png)

In #1241 I asked that the value be dynamic so that content creators can change the table header value. I would say we also change the "Notes" to "Description" or "Field Description" to better describe it's purpose.

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1274, opened 2021-01-11
Dynamic table header and displaying notes/description

sumandari:
This PR refers to #1241 and #847. It does:
- change the static requirement header table element into a dynamic value
- display note/description on printed material

---
web view

1241_web

---
PDF output

1241_pdf

---
Edit worksheet detail

1241_edit

---
Edit specification detail 

1241_edit1


zacharlie:
I think this is great 👍 , but I am a bit concerned about the labels of "notes" and "description". Both seem to perform the same function, but are applied to different columns. I would advocate that they be labelled **"Title Notes"** and **"Value Notes"**.

I don't know how important this is to change in the back-end. Personally I feel like just changing the form labels should be sufficient for now, as long as the function of these fields (and the labelling conventions) is properly documented so that other developers/ administrators might make more sense of this.

timlinux:
LGTM once @zacharlie 's comments factored in

sumandari:
Hi @zacharlie, I updated the PR:
- rename notes and description to title_notes and value_notes
- remove css styling for value_notes and title_notes so that it will have the same font-size with the title and the value as per @timlinux feedback.
  
image

image

zacharlie (approved):
LGTM

Issue #1257, opened 2020-12-28
Fix module number error after editing lesson content


zacharlie:
If a lesson is edited and saved, the page does not show the module numbers in the html or pdf.

![image](https://user-images.githubusercontent.com/64078329/103202300-b4f65b00-48fa-11eb-97b4-f562c4719a54.png)

Refreshing the page does not fix it, but navigating back to the module list and returning to the same module updates these values. This should happen automatically.

sumandari:
Hi @zacharlie, the fix #1275 has been deployed on staging server.

PR #1275, opened 2021-01-11
Persist section numbering after update the lesson

sumandari:
This PR refers to #1257. It does:
- add query string on edit page so that when it's submitted will display the numbering

changelog_1257_numbering


dimasciput (approved):

timlinux:
LGTM

Issue #1235, opened 2020-12-21
Change "Further reading" styles


zacharlie:
Currently, the Further reading elements are simply a space separated element with a text description and a hyperlink. This should be re-evaluated so that elements are clearly defined, legible and usable across various media types (digital/ print etc.)

At this time, I would be happy if the field was a simple rich text field that directly renders markdown/ html in order to offer flexibility to content creators. Alternatively, simply rendering the elements as a table may be a suitable approach.

**Note that the implementation of this has very serious considerations especially with regard to existing content and the possibility of data loss.** It is imperative that team discussions cover the design and implementation of this functionality using all the available communication channels.

sumandari:
Hi @zacharlie, just to confirm my understanding, do you mean to replace the text description and hyperlink field with a rich text editor (one field only)? Or only replace the text description field to a rich text editor?

I gave a try to use tinymce for text description editor. I was not sure if it should support html or markdown or both.
![changelog_further_lesson_editor](https://user-images.githubusercontent.com/40058076/103327204-95903700-4a8e-11eb-96f3-c472efdeeefd.gif)

zacharlie:
I would personally be happy with a single rich text field (preferably supporting both html and markdown, or promote markdown to html once posted), which would give us full control of the rendering for the web and print versions. I'm not entirely sure what implications that has for exposing the values via the models or API though so I think we would need input from @timlinux and @NyakudyaA to identify the best design.

For web versions, I feel that a single hyperlink is suitable, whereas the full URL and descriptor should be available in print versions. I think this should be as a table or similar format because the space separated list isn't very neat.

timlinux:
Happy to make this change, but @sumandari you will need to figure out a migration strategy for the old data.

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1277, opened 2021-01-13
Remove link field in furtherreading model, concatenate the old data in custom migration

sumandari:
This PR refers to #1235. It does:
- replace the text field editor and remove link field into a single rich text editor
- handle the old data in custom migration, concatenate link value with text value in text field.

changelog_1235_rich_tect_editor


timlinux:
LGTM

Issue #1245, opened 2020-12-22
Review strategy for worksheets


NyakudyaA:
# Problem

The lesson app is great but we need a way to review exercises between ourselves or external content creators. 

The current scenario is 
* John writes an exercise sheet and saves it. Even before finishing it becomes available for everyone to see it.

# Proposed Solution 

* John writes an exercise. He should save it and continue editing until he feels it needs a review from someone else. Whilst he is editing his worksheet at any time it should never be visible on the site to the public users.
* Foo user then goes to the site and check which exercises need reviewing. Once he is happy with it he then published the worksheets. ( In the future if we have more contributors writing exercise we should have email notification to let a content reviewer that there is material to be reviewed)
* An Admin user can still write material and publish it without having anyone to review it.

cc @zacharlie 
**NB** We currently have to this process manually to indicate to one another that can review my work.

zacharlie:
I think this seems like quite a technical challenge to implement the workflow you describe, so we should either discuss an alternative approach, and incremental implementation plan or a comprehensive workflow.

From my understanding you are proposing the following workflow:

Each worksheet has a "published" copy and a "draft" copy. When the edit button is pressed it navigates to the "draft" version.

We then need actions exposed for the following items:

- Approve Worksheet: copies the state from "draft" to the "published" object.
- Reset Worksheet: copies the state from "published" to the "draft" object.
- Request Review: Send email notification to reviewers. Each worksheet would need designated reviewer accounts which defaults to all content editors group, but can be modified to a specific user account or email address when the action is triggered.
- Preview: The application front-end only reads from the "published" object. When signed in as an editor, a "preview" option should be available for reviewing the rendered results from the "draft" copy. This should not be accessible by anonymous users. When editing the "draft" copy, saving the changes should default to the "Preview" page. The preview should include the export to PDF option as well.
- Review master: When on the preview page, editors should be able to quickly switch back to the master "published" object view

Ideally, If this intended for the long term with multiple contributors, I think this should be extended to the following:

The main worksheet state that is publicly viewable is retained as the master copy, which is made available after an approval/ review workflow (e.g. approved by a "content-admin" role).

An alternative, private, copy is retained as the latest "published" copy. If this get's approved it's state supersedes the master document. Until it get's given this approval, the review document is kept in a private state, and authors can request review from content administrators (group/ individuals) which sends email notifications.

I'm not sure whether this "review" copy should have a single instance, or a multi (per user) instance.

[...]

PR #1281, opened 2021-01-18
1245 unpublished worksheet

sumandari:
This PR refers to #1245. It does:
- add published field in worksheet models
- in normal view:  
  - display only published worksheet on section page
  - return 404 for unpublished worksheet detail view
- in admin view: 
  - display unpublished worksheet with `draft` mark in list view and detail view.

---
Left side is the admin/owner view, right side is the normal view.

image

---
numbering will change after published worksheet

image

---
detail view

image

---
publish the worksheet

image


NyakudyaA:
@sumandari if someone manually enters a URL for the worksheet that is not published I think you should redirect them to the login page rather than showing a 404. This is safer as it gives the indication that the worksheet is only visible to admins

sumandari:
Thanks @NyakudyaA, I updated the PR. It will redirect to the login URL. If user is anonymous, it will go to login page. If user already logged in but not an admin/owner it will go to home page.
changelog_1281_redirect_login

NyakudyaA:
LGTM

timlinux:
LGTM

Issue #1295, opened 2021-02-11
Unable to move lessons from one section to another


NyakudyaA:
# Problem

I am trying to move some lessons from 

* 1. Introduction to QGIS  to another section

i.e
15. Services and system administration 

Currently, there is no way to achieve this.

Can we add an option that allows admins to change worksheet sections

sumandari:
The fix has been deployed. I will close this now.

PR #1298, opened 2021-02-15
enable change section when update worksheet

sumandari:
This PR refers to #1295 
It enables creator/admin to change the section while updating the worksheet.

image


NyakudyaA:
LGTM

Issue #1004, opened 2019-02-15
Download multiple items Is limited to worksheets appearing on single page


NyakudyaA:
# Problem

I tried to download multiple sheets and I got to this page.

![screenshot 2019-02-15 at 15 52 23](https://user-images.githubusercontent.com/2510900/52860993-bfaa0b80-3139-11e9-9380-25a1c9d5af96.png)

When I navigate to the next page for worksheets it does not remember the state I was in of downloading multiple worksheets the previous state of downloading worksheets.

![screenshot 2019-02-15 at 15 54 20](https://user-images.githubusercontent.com/2510900/52861066-fc760280-3139-11e9-9ef7-5013cb87226b.png)
 
# Solution
Fix so that a user can download worksheets across various pages.

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1301, opened 2021-02-17
support pagination for 'download all worksheet' feature

sumandari:
This PR refers to #1004.

It enable select the worksheet across page.

changelog_1004


NyakudyaA (approved):
LGTM

Issue #1291, opened 2021-01-21
Neater links to external data in the pdfs


zacharlie:
Currently, the pdf links to external data/ exercise data don' render very nicely due to the archive name being a hash value and the link being very long. Can we try neaten this up? Thing I can think of to help would be to put the link on a new line, maybe wrap it in a span element, reduce the font size and possibly name the archive something shorter, e.g. `qgis1.1.zip`, where qgis is project name/ is and 1.1 is the module number.

![image](https://user-images.githubusercontent.com/64078329/105402072-a23c2100-5c2f-11eb-8aa6-47c04a39598f.png)

sumandari:
@zacharlie  do you mean we need create a new (shorten) url that redirect to this link ?

PR #1302, opened 2021-02-17
url link for lesson sample data

sumandari:
This PR refers to #1291.
It provides neater link for download external_data file
`/<project_slug>/sample/<worksheet id>?q=<module number>`

The url cannot use module number as an argument (as mentioned in the ticket), since the module number is not retrieved from the model.
The pdf output name is using project and module number as the filename. 

image
image

### DIFF
--- a/django_project/lesson/views/worksheet.py
+++ b/django_project/lesson/views/worksheet.py
@@ -587,6 +587,6 @@ class WorksheetSampleData(View):
             zip_file,
             content_type="application/x-zip-compressed")
         response['Content-Disposition'] = \
-            'attachment; filename="%s%s%s"' % (
+            'attachment; filename=%s%s%s' % (
                 project.name, numbering, extension)
         return response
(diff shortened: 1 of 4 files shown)


dimasciput (approved):

django_project/lesson/views/worksheet.py
```diff
@@ -564,3 +570,23 @@ def get_context_data(pk):
         'attachment; filename={}-worksheet {}.zip'.format(
             project.name, downloaded_module)
     return zip_response
+
+
+class WorksheetSampleData(View):
+    """Download sample data."""
+
+    def get(self, request, *args, **kwargs):
+        numbering = self.request.GET.get('q', '')
```
dimasciput, line 579: can you get this version number from the project model? so you don't need to pass this variable to view because it seems redundant.

sumandari: The numbering is not saved in the model instance.
It's coming from dynamic forloop in the template. 
https://github.com/kartoza/prj.app/blob/58aec98309b3aa1ca18b9c776e5b9bac2a086126/django_project/lesson/templates/section/includes/row-worksheet.html#L12
and the numbering might be different between author and normal user, since the author have view in draft (unpublished) worksheet, but the normal users don't.

timlinux: LGTM

Issue #1317, opened 2021-02-24
Fix GitHub Import


sumandari:
Can you also fix the GitHub importer so that when the images are initially imported they are correct?

_Originally posted by @timlinux in https://github.com/kartoza/prj.app/issues/1306#issuecomment-782080750_

Gustry:
Duplicate of https://github.com/kartoza/prj.app/issues/1230 ?

sumandari:
@Gustry this is to avoid/fix issue #1304 and #1305.
And yes, #1305 is duplicate of #1230. Thank you for pointing it out.

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1318, opened 2021-02-24
fix github importer and update unit test

sumandari:
This PR refers to #1317.

It fixes the url path and update the unit test.


sumandari:
Travis gave prefix `/tmp` in image path
it causes the test failing when my local test passed

sumandari:
please review @lucernae @dimasciput. Thank you

django_project/changes/tests/test_github_pull_request.py
```diff
@@ -236,10 +249,32 @@ def test_download_all_referenced_images(self):
...
+        self.assertTrue(entry.image_file)
+
+        self.assertIn(
+            '106831433-dea95b80-66ca-11eb-8026-6823084d726e',
+            entry.image_file.name
+        )
+
+        image_file_dirname = os.path.dirname(entry.image_file.name)
+        image_file_upload_to = re.sub(r'^/', '', image_file_dirname)
+        image_file_url = settings.MEDIA_URL + image_file_upload_to
```
lucernae: It's better if you define a constant in `Entry` model class with the relative location of `image_file_upload_to`. Then, there's no need to check `field.upload_to` below, CMIIW, because both came from a constant.

for `image_file_url` constructions, use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin 

when comparing url below, extract only the urlpath (without hostname) to compare.

django_project/changes/tests/test_github_pull_request.py
```diff
@@ -236,10 +249,32 @@ def test_download_all_referenced_images(self):
...
+        self.assertEqual(
+            entry.image_file.field.upload_to,
+            image_file_upload_to
+        )
+
+        # check if upload_to give the same path
+        self.assertEqual(
+            os.path.dirname(entry.image_file.url),
+            image_file_url,
+        )
```
lucernae: Do not use `os.path.dirname` for url operation because it can be misleading for the reader.
Use urllib.

Also, I'm not sure why you want to compare only the dirname and not the image name too.
However if `upload_to` is using a constant (declared in the `Entry` model class), it can be guaranteed that it will be the same, so you don't need to test it like this, but rather the full path.

django_project/changes/views/changelog_github.py
```diff
@@ -260,13 +260,18 @@ def download_all_referenced_images(request, **kwargs):
                             folder_path, '{}'.format(filename))
                         found = os.path.exists(file_path)
                         if found:
+                            # file_path
+                            # e.g /home/web/media/images/entries/img.png
```
lucernae, line 279: Nice. It's always good to put some comments.

django_project/changes/views/changelog_github.py
```diff
@@ -260,13 +260,18 @@ def download_all_referenced_images(request, **kwargs):
...
                             while found:
                                 n += 1
                                 img_name = file_path_original.rsplit('.', 1)[0]
                                 extension = \
                                     file_path_original.rsplit('.', 1)[-1]
+                                # create a unique filename:
+                                # add sufix -n in filename prior to extension
+                                # e.g /home/web/media/images/entries/img-1.png
                                 file_path = \
                                     img_name + '-' + str(n) + '.' + extension
```
lucernae: Use Python's f-string.

django_project/changes/views/changelog_github.py
```diff
@@ -260,13 +260,18 @@ def download_all_referenced_images(request, **kwargs):
...
                         if found:
+                            # file_path
+                            # e.g /home/web/media/images/entries/img.png
                             file_path_original = file_path
                             n = 0
                             while found:
                                 n += 1
                                 img_name = file_path_original.rsplit('.', 1)[0]
                                 extension = \
                                     file_path_original.rsplit('.', 1)[-1]
```
lucernae: Do not split twice, but rather:
This works because the split only returns two-tuple, so it's easier to read.

```suggestion
                                img_name, extension = file_path_original.rsplit('.', 1)
```

lucernae: But actually we should rather use `os.path.splitext` for this kind of things.

django_project/changes/views/changelog_github.py
```diff
@@ -280,14 +285,23 @@ def download_all_referenced_images(request, **kwargs):
...
                                     break
                                 handle.write(block)
-                            img_url = file_path.replace('/home/web', '')
+
+                            # remove MEDIA_ROOT to obtain image_file relative
+                            # path and add the MEDIA_URL path into the file
+                            # path e.g img_url: /media/images/entries/img-1.png
+                            img_url = file_path.replace(
+                                settings.MEDIA_ROOT,
+                                re.sub(r'/$', '', settings.MEDIA_URL))
```
lucernae: Use, `os.path.relpath` instead to get the relative path of the image based on `settings.MEDIA_ROOT`.

Then use urllib to construct the url.

django_project/changes/views/changelog_github.py
```diff
@@ -280,14 +285,23 @@ def download_all_referenced_images(request, **kwargs):
...
+                        # it from the body
                         if i == 0:
-                            # Take the first image set in the pull request
-                            # and set it as default for the entry and remove
-                            # it from the body
-                            entry.image_file = img_url.replace('/media', '')
+                            # remove MEDIA_URL and assign the value to
+                            # image_file
+                            entry.image_file = img_url.replace(
+                                settings.MEDIA_URL, '')
```
lucernae: It doesn't make sense because the right side is url but the left side is image field, CMIIW?
Why not use the `file_path` ?
Or better way, use the file stream as input to `image_file.save` like in this article: https://nikhilhaas.com/downloading-and-saving-image-to-imagefield-in-django/

Check how they uses `cStringIO` to stream the download, then pass it as `ContentFile` to image_field's `save`.

django_project/changes/tests/test_github_pull_request.py
```diff
@@ -29,11 +32,13 @@ def __init__(self, json_data, status_code):
         def json(self):
             return self.json_data
 
+    if args[0] == 'https://api.github.com/users/qgis':
+        return MockResponse({
+            'html_url': 'https://github.com/qgis',
+            'name': 'name'
+        }, 200)
```
lucernae, line 38: Explain in comment, why you need to do an if switch if the argument is like that?
Why the html_url is different too.

lucernae (commented):
Sorry for the long review.
Try to improve the test files first as priority, then implementation can follow as long as we are testing the right thing.

django_project/changes/tests/test_github_pull_request.py
sumandari: Thank you @lucernae for reviewing. I will partially update this PR since I need to digest the review and learn the solution as per your suggestion :sweat_smile:  

> Also, I'm not sure why you want to compare only the dirname and not the image name too.

when I do test on my localmachine, the image name will be
```images/entries/106831433-dea95b80-66ca-11eb-8026-6823084d726e-148.png``` 
instead of 
```images/entries/106831433-dea95b80-66ca-11eb-8026-6823084d726e.png```
 
There will be suffix `-n` in filename prior to the extension file comes from 
https://github.com/kartoza/prj.app/blob/dbf642b9185448f582fd42d3c1f28a81336b1806/django_project/changes/views/changelog_github.py#L271

sumandari: so that I don't know what the value should be compare with filename since the `-n` will give int number that I don't have idea how to address it.

django_project/changes/tests/test_github_pull_request.py
sumandari: ah, yes. this one used for 
https://github.com/kartoza/prj.app/blob/dbf642b9185448f582fd42d3c1f28a81336b1806/django_project/changes/tests/test_github_pull_request.py#L147

I was thinking to use mock during the test, I changed the return value and arguments so that it can be utilized by another test. haha, it ended up, i didnt use it in this PR.

Ya, comment is very important to explain my code :sweat_smile: . Thanks Rizky.

sumandari:
please kindly review the update @lucernae. Thank you!

sumandari:
cc @lucernae please kindly review when you get a chance, there's another issue related this:  #1348. Thank you

lucernae (approved):
LGTM! Thanks @sumandari

PR #1332, opened 2021-04-20
Fix the name encoding for sample data in lessons

### LINKED ISSUE #1329: Fix the name encoding for sample data in lessons
# Problem

Currently, all sample data is being downloaded with some encoding issues

i.e 
adding-vector-data

This has been fixed for the content in the sheets but we need to fix the names as well

sumandari:
This PR fixed #1329 

### Before:
image

---

### After:
image


NyakudyaA:
@sumandari an we get this merged as well

Issue #1337, opened 2021-05-03
Add Download PDF and Data all in one button in worksheet


timlinux:
On the lessons list we have this button to download the lesson pdf and data in one zip

![image](https://user-images.githubusercontent.com/178003/116856986-f32ac080-abf3-11eb-9199-ae922c681959.png)

Can we have the same button shown at the top of the worksheet too:

![image](https://user-images.githubusercontent.com/178003/116857075-19e8f700-abf4-11eb-9412-93e353f222e2.png)

Also when the worksheet + data zip downloads, it has another zip inside it with the data. Can you please unpack that so that the user does not need to unzip the first zip then go inside and unzip the second?

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1338, opened 2021-05-03
added download button on worksheet detail

sumandari:
Please see #1337 

image

### DIFF
--- a/django_project/core/base_static/img/pdf_download.svg
+++ b/django_project/core/base_static/img/pdf_download.svg
@@ -1,0 +1,112 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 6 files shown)


zacharlie:
Not sure if it matters whether the download data button should come before or after the PDF button. My thinking would be to move the download data button before the PDF button because not all lessons include sample data and it may remain a bit more consistent. maybe @timlinux can advise.

timlinux:
> Not sure if it matters whether the download data button should come before or after the PDF button. My thinking would be to move the download data button before the PDF button because not all lessons include sample data and it may remain a bit more consistent. maybe @timlinux can advise.

Yup that makes sense thanks - @sumandari will you do that? Will be great if you can get this deployed tomorrow @sumandari before training starts...

dimasciput:
Hi @sumandari , can you use these icons for the download buttons
[download_icon.zip](https://github.com/kartoza/prj.app/files/6418603/download_icon.zip)

dimasciput (approved):
LGTM

sumandari:
This fix has been deployed on changelog.kartoza.com and changelog.qgis.org version 2.1.5

timlinux:
Thank you very much @sumandari !

Issue #1140, opened 2020-02-17
Email certificate link is not working


spatialthoughts:
I held a certification workshop recently and issued certificates. When I click the 'email participants with a link to their certificates' it says success, but participants are not receiving any email. 

Can you please look into this?
https://changelog.qgis.org/en/qgis/certifyingorganisation/spatial-thoughts/course/qgis_custimizing-qgis-with-python_2020-02-15-2020-02-15/

timlinux:
@ann26 can you look at this please?

@spatialthoughts keep filing those issues if you hit problems - we will fix as quickly as we can. Thanks!

spatialthoughts:
Update: It seems the messages are being delivered but they are detected as Spam. See attached screenshot. Also name, organization name fields are byte strings with a 'b' prefix that should be converted to a string for the email

![Screenshot_20200219-073201](https://user-images.githubusercontent.com/5227506/74806971-542ae900-530d-11ea-811c-1d143f7c10a2.png)

I didn't realize the certificates are sent from the issuing organization's email. It will be better and preferable for the certificate email to come from qgis.org email. That looks more professional and can also avoid emails going to spam.

timlinux:
Yes this also relates to #1144 - the advertised domain should be consistent everywhere.

nickbearman:
+1 
Also having this issue with certificates that I have sent out. The participants claimed not to have recieved any emails (even in SPAM).

timlinux:
@sumandari can you look into this please?

nickbearman:
Just FYI, I have had one more person confirm they haven't recieved the certificate email, and it is not in their spam.

sumandari:
@nickbearman thank you for your report. Could you please inform the detail, which course the attendees haven't received the certificate email?
I have looked into the log error and haven't found any issue related email recently. Your detail information will help to identify the issue.

nickbearman:
Sure @sumandari 
It is the Introduction to QGIS: Spatial Data and Spatial Analysis | April 20, 2021 under Geospatial Training Solutions. 
https://changelog.qgis.org/en/qgis/certifyingorganisation/geospatial-training-solutions/course/qgis_introduction-qgis-spatial-data-spatial-analysis_2021-04-20-2021-04-28/

Participants impacted: (all UK time)
Juan Manuel Moreno, sent certificate through online system 10:14 30th April, user confirmed not recieved on 5th May 09:45, including checking spam. 
[...]

PR #1340, opened 2021-05-11
Update email setting

sumandari:
Please see #1140


dimasciput:
@sumandari any update on this?

sumandari:
@dimasciput  The gmail account setup for the related issue has not been completed yet.

Hi @timlinux, would you please advise whether we will continue to use the gmail-api or we will stick with the existing smpt? Thank you

timlinux:
We can use the gmail API. Can you get the pr running on a local instance so that we can do a test before deploying live.

Issue #1331, opened 2021-04-13
Elegant aspect ratios for youtube embeds


zacharlie:
Youtube video embeds have different aspect ratios and the lessons video frame is square, resulting in a lot of negative space surrounding the video.

![image](https://user-images.githubusercontent.com/64078329/114528851-aa619680-9c49-11eb-8843-af9c51a0d7dc.png)

A couple of resources describing the issue and possible fixes include:

* https://stackoverflow.com/questions/13880745/embedding-youtube-videos-aspect-ratio
* https://www.ankursheel.com/blog/full-width-you-tube-video-embed
* https://tubularlabs.com/blog/youtube-aspect-ratio/

Unfortunately these suggestions seem to assume that the video content will have the same dimensions, which is not necessarily the case and ideally this frame should dynamically adjust to suit the content.

If having to make it static, 16:9 (typical screen ratio) is probably best, but really, you would think that at this point an elegant solution exists for dynamic video content, so if you can find one please implement it. We definitely do not want to crop or distort any content either, so if we have to live with the negative space, so be it.

sumandari:
I was following the steps on https://www.ankursheel.com/blog/full-width-you-tube-video-embed, modifying the CSS, but it didn't work as expected.
On #1343 instead of changes the YouTube ratio, it changes the iframe ratio. I'm not sure if it's required.

PR #1343, opened 2021-05-17
Set embedded youtube player aspect ratio to 16:9

sumandari:
This is to address #1331 

1331_youtube


dimasciput:
@sumandari can you update me on this draft?

sumandari:
@dimasciput 
The ticket requires to set ratio youtube ratio video embedded on the web page. 
The solution I made in this draft is only to resize the frame size with a ratio 16:9, it doesn't work on the embedded youtube video itself.
I think I will close/decline this PR since it's not relevant to the issue

Issue #1345, opened 2021-05-24
Anchor tag in Further reading editor


sumandari:
When added the anchor tag, the tinyMCE does save it as plain text.

![image](https://user-images.githubusercontent.com/40058076/119284096-78444b00-bc71-11eb-8b0a-0ceb17585c05.png)

![image](https://user-images.githubusercontent.com/40058076/119284107-809c8600-bc71-11eb-894b-394957b4aa41.png)

PR #1346, opened 2021-05-24
Fix Further reading menubar editor

sumandari:
To fix the TinyMCE in lesson further reading editor #1345 

image

image


dimasciput:
@sumandari can you create a single html template to initiate the tinymce instead? So if you need to update the configuration, you only need to do it once.

dimasciput (approved):
LGTM

Issue #1292, opened 2021-02-04
Allow option to highlight correct answers for webview exercise


NyakudyaA:
# Problem

When I am doing the lessons I cannot determine which is the correct answer to a question.
![answer](https://user-images.githubusercontent.com/2510900/106864448-8c6c3880-66d2-11eb-9a22-4029483a89a5.png)

# Solution

Highlight the correct answer when a user hovers or clicks on the answer. Alternatively, you can add the option for the user to submit the answers and then show the correct answer

christopherbuirski:
![Solution ](https://user-images.githubusercontent.com/78072657/106900273-00bcd100-66ff-11eb-9fad-78d39519f8b1.jpg)

sumandari:
The fixed has been deployed, I will close this ticket now

PR #1350, opened 2021-06-11
Added highlight on answers exercise

sumandari:
This is to address #1292 

- The highlight feature will be available to the authenticated user (and admin).
- Disabled the click event when answers have been submitted.

1292_highlight_answer


NyakudyaA:
LGTM @sumandari , Can we get a tooltip on the `Check your knowledge Icon` to let users know they can click on the answer to select their choice or is it already catered for somewhere

sumandari:
Added Tooltip:

image

PR #1358, opened 2021-07-23
Update slug to be unique together with certifying organisation

### LINKED ISSUE #1357: [Certification] Error creating a new course
I have been trying to create a new course under Septima's certification page and I get this error 500:

> 500
Ooops. Something broke....

>The powers that be have been informed.
In the meantime, please form an orderly queue and head to the home page.

>If you need assistance, you may reference this error as 77ea4aea60af42e087c4ac16108477f1.

I created a course here successfully this spring so I am not sure what is causing this.

sumandari:
This is to fix #1357 

When user create new course type, it will create slug with 50chars max from course name.

This PR enable the different certifyng organisation to create the same course name.


sumandari:
Notes:

- When user create a new course type, it will create a slug with 50chars max from **CourseType `name`**. If the course name is **longer than 50 chars**, it will lead to a slug duplication.
- We have applied a mechanism to check and add counter after`name` when it exists and create a slug from it. But the unique field is `slug` instead of `name`.

In the issue #1375, there is an existing courseType with same `slug` but it has different `name`.
I guess the existing courseType was edited with a new name, and that's why our code cannot prevent this duplication.
So I think we can change the way we add counter in the slug duplication rather than change the unique value.
Would you have suggestion @dimasciput?

sumandari:
We are not implementing this. I will close this PR

PR #1361, opened 2021-08-07
Add github action to test PR

sumandari:
This PR is using docker-compose to test against PR


sumandari:
To test if the test against PR failed:
`test_always_failed.py`
image

.github/workflows/build-test-pr.yaml
```diff
@@ -0,0 +1,41 @@
...
+      PGDATABASE: docker
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build containers
+        run: docker-compose --file deployment/docker-compose.test.yml build
+      - name: Run db container
+        run: docker-compose --file deployment/docker-compose.test.yml up -d db
+      - name: Run devweb container
+        run: docker-compose --file deployment/docker-compose.test.yml up --no-deps -d devweb
```
lucernae: You can reuse the docker build in action in https://github.com/kartoza/prj.app/blob/develop/.github/workflows/build-push-images-latest.yaml .

.github/workflows/build-test-pr.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Run coverage test
+        run: docker-compose -f deployment/docker-compose.test.yml exec -T devweb bash -c 'coverage run manage.py test --settings=core.settings.test_docker'
+      - name: Run coverage report
+        run: docker-compose -f deployment/docker-compose.test.yml exec -T devweb bash -c 'coverage xml'
+      - name: Upload coverage to codecov
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Run flake8
+        run: docker-compose -f deployment/docker-compose.test.yml exec -T devweb bash -c 'flake8 .'
```
lucernae: This is a good approach. Running the test inside the container ensures the same environment.

Just need some improvements:
- extract test parameters as environment variable
- I'm guessing coverage.xml is inside the container. You need to copy it out.
- flake8 should have run first. (we check the code convention first. If linting works, then run unittest/coverage)

See the original .travis.yaml file to understand the test flow.

deployment/docker-compose.test.yml
```diff
@@ -0,0 +1,50 @@
+version: '3'
```
lucernae, line 1: The proper way is to set environment variables, or `.env` file for testing, and then use the override file in `docker-compose.override.example.yml`. We should delete this file if possible.
We can use the file, but only if it extends the original `docker-compose.yml` not replace it.

lucernae (commented):
I can submit a patch to demonstrate the fix. You can then apply it to other project.

.github/workflows/build-test-pr.yaml
sumandari: Does it mean, I can edit the docker-file and using this image instead 
`${{ secrets.DOCKERHUB_REPO }}/projecta-uwsgi:latest` 
and I pass the secret value to env to be used by docker-compose.test ?

lucernae:
Merge my PR https://github.com/sumandari/prj.app/pull/10 first so that I can back-review it in this PR.

.github/workflows/build-test-pr.yaml
lucernae: Ideally the deployment should be configurable from just a config file or environment/parameters, so that the deployment code can be reused.

The plan is like this:
- Specify files/environment/parameters for running unittests
- Use docker build action to build image
- Use that image for the unittest
- If succeed, rebuild the image from cache, but this time, uses proper docker tag and then push to image registry.

To answer your question:

> Does it mean, I can edit the docker-file and using this image instead
`${{ secrets.DOCKERHUB_REPO }}/projecta-uwsgi:latest`
and I pass the secret value to env to be used by docker-compose.test ?

Use the docker build action to build the image. Since this is for test (we won't push it), you can just replace `${{ secrets.DOCKERHUB_REPO }}/projecta-uwsgi:latest` with hardcoded name, or Github Action env parameter. For example: `kartoza/projecta-uwsgi:test`. There is no need to pass the secret value to docker-compose.test since the image is already been built and available in the machine. You just need to use the matching tag in `docker-compose.test.yml`. 

Also, it's better to override instead of using `docker-compose -f docker-compose.test.yml`. Copy `docker-compose.test.yml` as `docker-compose.override.yml`. If you run `docker-compose up -d` it will automatically merged the yaml files of both `docker-compose.yml` and `docker-compose.override.yml`.

.github/workflows/build-push-images-latest.yaml
```diff
@@ -5,14 +5,111 @@ on:
     branches:
       - develop
 jobs:
+  lint:
```
lucernae, line 8: This file is very similar with `.github/workflows/build-push-images-pr.yaml`.

The main difference is just it always push the docker image in `docker-builder` job.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+        uses: actions/setup-python@v2
+
+      - name: Test installing development dependencies
+        working-directory: deployment/docker
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r REQUIREMENTS-dev.txt
+
+      - name: Run Flake8 test
+        run: flake8 .
```
lucernae, line 26: Lint job always runs first.
Avoid building anything for linting, so when it fails, it fails quickly.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+
+  test:
+    runs-on: ubuntu-latest
+    needs:
+      - lint
+    env:
+      APP_IMAGE: kartoza/projecta-uwsgi
```
lucernae, line 33: I'm not sure why there are `core.settings.dev_docker.py` and `core.settings.test_docker.py`. Since we have two settings, the local image build need to have this `APP_IMAGE` parameter, so that you can create docker-compose override that use this parameter.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+          context: .
+          file: deployment/docker/Dockerfile
+          push: false
+          load: true
+          target: dev
+          tags: ${{ env.APP_IMAGE }}:dev
+          cache-from: |
+            type=gha,scope=test
+            type=gha,scope=prod
+          cache-to: type=gha,scope=test
```
lucernae, line 53: This step does:
- Building the APP_IMAGE with `dev` tag for testing
- Load the image to the local docker service
- Set build target to be `dev` (corresponds to the target name in Multi Stage Dockerfile)
- Cache image to Github Cache Action, with scope=test
- Retrieve cache image from Github Cache Action if exists, with priority scope=test, if not exist, scope=prod

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+          context: .
+          file: deployment/docker/Dockerfile
+          push: false
+          load: true
+          target: dev
+          tags: ${{ env.APP_IMAGE }}:prod
+          cache-from: |
+            type=gha,scope=test
+            type=gha,scope=prod
+          cache-to: type=gha,scope=prod
```
lucernae, line 67: Similar with above, but now builds the `prod` target as `prod` tag.
It will reuse previous cache above because building `dev` target also build `prod` target as dependencies.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+
+      - name: Run docker-compose services
+        working-directory: deployment
+        run: |
+          echo "Override docker-compose for testing purposes"
+          cp .env.example .env
+          cp docker-compose.test.yml docker-compose.override.yml
```
lucernae, line 74: Just copy exisiting `.env.example` as `.env` to define the docker-compose parameter.
Also copy the `docker-compose.test.yml` as the `docker-compose.override.yml`

Regular docker-compose command will use these files by default without any extra command flags.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+          make up
+          make devweb
+          make wait-db
+          make create-test-db
```
lucernae, line 78: We already have some command in the `Makefile`, just reuse it here instead of executing `docker-compose` again.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+      - name: Run Coverage test
+        working-directory: deployment
+        run: |
+          cat << EOF | docker-compose exec -T devweb bash
+          python manage.py makemigrations
+          python manage.py migrate
+          python manage.py collectstatic --noinput --verbosity 0
+          coverage run manage.py test
+          coverage xml
+          EOF
```
lucernae, line 89: Piping with heredoc. The command itself is:

```bash
python manage.py makemigrations
python manage.py migrate
python manage.py collectstatic --noinput --verbosity 0
coverage run manage.py test
coverage xml
```

It is passed over to the bash interpreter inside the container. It is executed inside the container.
`-T` flag is used to ignore TTY.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+          docker cp projecta_devweb_1:/home/web/django_project/coverage.xml ../coverage.xml
```
lucernae, line 90: This is what I meant in the previous review. You need to copy the `coverage.xml` file out of the container.

We run the services without bind-mount to `django_project` to also make sure that we are indeed using the code **inside the container**. If the unittest passed, then the we are very sure the docker image is healthy.

If we bind mounted `django_project` inside the container, we are not testing against the code inside the image.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+
+      - name: Upload coverage to codecov
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        if: env.CODECOV_TOKEN != ''
```
lucernae, line 95: Contributor fork might not have CODECOV integration, so we skip this step in that case.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+        uses: codecov/codecov-action@v2
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          fail_ci_if_error: true
```
lucernae, line 99: Must fail the step if codecov doesn't push.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -4,15 +4,113 @@ on:
...
+
   docker-builder:
+    # Only push if PR happens in the same repo
     if: github.event.pull_request.base.repo.url == github.event.pull_request.head.repo.url
     runs-on: ubuntu-latest
+    needs:
+      - test
```
lucernae, line 106: This step is executed if:
- The test success (the previous job)
- Push to our kartoza registry if PR happens between two branches in the same fork.

.github/workflows/build-push-images-pr.yaml
```diff
@@ -34,6 +134,13 @@ jobs:
           context: .
           file: deployment/docker/Dockerfile
           push: true
+          target: prod
           tags: ${{ steps.docker_meta.outputs.tags }}
+          labels: ${{ steps.docker_meta.outputs.labels }}
+          cache-from: |
+            type=gha,scope=test
+            type=gha,scope=prod
+          cache-to: type=gha,scope=prod
```
lucernae, line 143: We only push the `prod` target image.

This step will reuse the previous cache in the previous job. We already build the image, so we only push this time.

codecov.yml
```diff
@@ -2,3 +2,5 @@ coverage:
   range: 50..75
   round: down
   precision: 2
+fixes:
+  - "/home/web/django_project/::django_project/"
```
lucernae, line 6: Coverage path inside container needs to be fixed to the repo path.

deployment/Makefile
```diff
@@ -50,6 +50,13 @@ db:
 	@echo "------------------------------------------------------------------"
 	@docker-compose up -d db
 
+wait-db:
+	@docker-compose exec -T db su - postgres -c "until pg_isready; do sleep 5; done"
+
+create-test-db:
+	@docker-compose exec -T db su - postgres -c "psql -c 'create database test_db;'"
+	@docker-compose exec -T db su - postgres -c "psql -d test_db -c 'create extension postgis;'"
```
lucernae, line 58: Helper to generate test database.

We don't actually need it because Django prepares the database, but since the test uses `core.settings.test_docker` where it specified the default database, then we have to set this.

deployment/docker-compose.override.example.yml
```diff
@@ -52,8 +53,9 @@ services:
   devweb:
     image: kartoza/project-uwsgi:dev-latest
     build:
-      context: docker
-      dockerfile: Dockerfile-dev
+      context: ${PWD}/../
+      dockerfile: deployment/docker/Dockerfile
+      target: dev
```
lucernae, line 58: Refactored the Dockerfile-dev to be merged inside Dockerfile file to use multi-stage build.

Specify the build target `dev`

deployment/docker/Dockerfile
```diff
@@ -29,3 +29,37 @@ EXPOSE 8080
...
+# --------------------------------------------------------
+# Open ports as needed
+# --------------------------------------------------------
+
+# Open port 8080 as we will be running our django dev server on
+EXPOSE 8080
+# Open port 22 as we will be using a remote interpreter from pycharm
+EXPOSE 22
+
+CMD ["/usr/sbin/sshd", "-D"]
```
lucernae, line 65: We only moved the content of `Dockerfile-dev` to `Dockerfile`, then:

- Modified the FROM command to use `prod` target as a base
- Adjust ADD parameter's path to be from the repo root directory

deployment/docker/build.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
-cd ../../
-docker build -t kartoza/projecta-uwsgi -f deployment/docker/Dockerfile .
+pushd ../../
+docker build -t kartoza/projecta-uwsgi -f deployment/docker/Dockerfile --target prod .
+popd
```
lucernae, line 4: Now we specify the target `prod`

This script is only used if you want to build the image manually (legacy reason).

A better way is to just use `docker-compose build`.

lucernae (approved):
Check the Review comment here to understand the changes.

.github/workflows/build-push-images-latest.yaml
lucernae: Maybe we can merge these files.

lucernae:
Ah we don't need any token at all if it is a public repo.
image

sumandari:
I have got this error once:
image
in upload codecov steps at https://github.com/sumandari/prj.app/runs/3268591845


> Ah we don't need any token at all if it is a public repo.
> image

----
Thank you @lucernae for your kind help, your valuable work and all details above.

PR #1362, opened 2021-08-11
Use Codecov convention for public repo

lucernae:
- We don't have to use CODECOV_TOKEN parameter to push to codecov
- Github Action for latest and PR is now merged


sumandari:
Mantab! 
Thank you @lucernae 
I can see the codecov report now

sumandari (approved):

PR #1367, opened 2021-10-01
added boostrap files to static

sumandari:
To fix the errors we got recently:
image


sumandari:
It seems related to ISP problem. I will close this PR for now.

Issue #1372, opened 2021-11-15
Cannot replace sustaining member logo


andreasneumann:
I am trying to replace the already existing logo from sustaining member "Automatic knowledge" (https://changelog.qgis.org/en/qgis/member/dvtftt/update/) with a new version - but every time I try to replace this particular logo, I get a 500 webserver error.

Is there something special with this particular entry?

Thank you for your help!

sumandari:
Hi @andreasneumann thank you for your report. 
From the sentry error, it found that your email address is not valid. Please kindly check your email domain in `Contact Email`. It probably needs to be updated.

And I think we should display the ValidationError on the web instead of return 500. I will look into this

andreasneumann:
Strange - If I go to my projecta account settings, everything looks normal

![image](https://user-images.githubusercontent.com/884476/141763498-24565463-9663-4b6b-9ed5-91b0cf93980b.png)

sumandari:
I meant the contact email in the `Update Sustaining Member for QGIS` page (https://changelog.qgis.org/en/qgis/member/dvtftt/update/):
![image](https://user-images.githubusercontent.com/40058076/141764014-4fd02fc2-dba8-4041-8e68-91bb9681f236.png)

andreasneumann:
Ah - yes. After fixing the invalid e-mail I was able to replace the logo.

Thanks for your help!

sumandari:
Cool!
And thanks again for your valuable report. It helps us improve the platform.

sumandari:
The fix has been released in v2.3.0 and deployed on prod.

PR #1376, opened 2021-11-26
Fix raising ValidationError in validate_email_address

sumandari:
Please see #1372. 

There was return 500 when user updated information in `Update Sustaining Member` page. After replicated the issue, I found that  fixing the way we raise ValidationError in validate_email_address will resolve this issue

image


timlinux:
LGTM

PR #1437, opened 2022-06-03
Update course attendee table

dimasciput:


dimasciput:
For https://github.com/kartoza/prj.app/issues/1433

PR #1439, opened 2022-08-31
GH PR parser - Use links in markdown for funders

Gustry:
It happens often. This test case should be covered.


Gustry:
Ah, tests are already passing ?
But the logs in the GH Action shows it failed !
https://github.com/kartoza/prj.app/runs/8111351705?check_suite_focus=true#step:8:932

@dimasciput This looks dangerous FYI.

dimasciput:
@Gustry thanks for noticing, I will check this issue when I'm back from leave. Did you make this PR while resting from your bike trip? 😄

PR #1452, opened 2023-11-29
add matomo metrics in base template

Akiyamka:


timlinux:
@Akiyamka note that there are more than one instance of the changelog site out there so we need to softcode the URL into the site configuration.

timlinux:
e.g. here https://github.com/kartoza/prj.app/blob/develop/deployment/.env.example

Akiyamka:
I looked at how it is done for intercom and assumed that it would be correct to do the same for matomo. Is that correct?

timlinux:
@dimasciput  could you review and deploy on qgis changelog please?

Xpirix:
@Akiyamka I think implementing django-matomo like you did in https://github.com/qgis/QGIS-Django/pull/320/files is a good way for django based projects. Thus, in the existing settings files, matomo variables can be loaded from the environments variables using `os.environ` that are specified in the `.env` generated from the sample `.env.example`.

Xpirix:
@Akiyamka The django-matomo template can be customized. To disable cookies, for example: https://github.com/qgis/QGIS-Django/pull/333

Issue #1461, opened 2024-05-23
Upgrade stacks and docker compose


Xpirix:
We plan to migrate the changelog app to the New QGIS Infrastructure. We can take advantage of this opportunity to update the application stacks and docker-compose services:
- [x] Upgrade to Python 3.12
- [x] Update requirements: upgrade to Django 4, upgrade other packages
- [x] Upgrade docker services: Postgis 16, Dbbackups 16
- [x] Fix various errors in the code after the upgrade (https://github.com/Xpirix/prj.app/issues/1)

PR #1462, opened 2024-05-23
Upgrade stacks

Xpirix:
This is for #1461


dimasciput (approved):

Xpirix:
Will be applied with #1503

Issue #1463, opened 2024-05-29
404 return when pressing "Generate all certificate for this course"


jeeemil:
I try to press the "Generate all certificate for this course"-button after I've added all the course participants through a csv file. When I try to generate all the certificates in one go, I get a 404 error page.

Generating certificates one by one works fine.

Is it possible that there's a broken link somewhere behind the "Generate all certificate for this course"-button? I couldn't find any other issue regarding this at the moment. Thankful for any advice.

jvdkwast:
I sometimes also experience this when I have many certificates. Maybe there's a time-out issue.

jvdkwast:
@Xpirix I just had it now (12:02 UTC) when I wanted to generate 21 certificates. Maybe you can find it in the log?
![image](https://github.com/user-attachments/assets/d516cc3a-c0b3-4653-b259-7847bf654797)

jvdkwast:
Note that it's a 502 and not 404 error.

Xpirix:
Thanks for reporting it @jvdkwast, I will try to find the origin of this issue.

PR #1469, opened 2024-09-24
Optimize nginx and uwsgi for generating certificates

Xpirix:
Fix for #1463 

- Increase timeout for nginx request
- Increase uwsgi harakiri to 100 (same as plugins.qgis.org)


dimasciput:
@Xpirix, thanks! This is fine for now. However, the ideal approach would be to handle this in the background to avoid timeouts. Perhaps consider integrating it with celery for future enhancements.

dimasciput (approved):

PR #1471, opened 2024-11-12
Allow course attendees without surname

### LINKED ISSUE #1459: Allow participants without a surname
Hi,

Currently the certification system doesn't allow adding participants without a Surname field. Many parts of the world only given names are used (see [Mononym](https://en.wikipedia.org/wiki/Mononym) )a nd thus it is not possible to issue a certificate without workarounds.

I would request a change in the system to name the Surname field optional to allow participants with a mononym to get a certificate

Xpirix:
Fix #1459 

- Make Surname optional
- Handle None value for Surname


dimasciput:
@Xpirix could you also fix the test? I think you need to change docker-compose to `docker compose` in Makefile.

Xpirix:
@dimasciput , the test workflow should be fixed along #1469. I could also submit a separate fix for the test, what do you think?

dimasciput (approved):

Issue #1380, opened 2021-12-01
URL in the certificate is showing HTTP without HTTPS


Gustry:
In the preview certificate (which is visible only in editing mode on the course, it's not very convenient, I would expect this button on the main page as well),
we can read

> You can verify this certificate by visiting http://changelog.qgis.org/en/qgis/certificate/QGIS-DEMO/.

which looks like a little bit surprising between "verify" and not being "https" ;-)

PR #1474, opened 2025-01-16
Fix http link in cerficates

Xpirix:
Fix for #1380

## Changes summary
- For new certificates: The link will be in `https`
- For existing certificates: They should be regenerated to have an `https` link


timlinux:
LGTM thanks

dimasciput (approved):

Issue #1366, opened 2021-09-08
Allow breakin the lines for the 'trained competences' field


andremano:
It would be useful to allow more than one line under the 'trained competences' field of a certificate. This would prevent situations like the example below: 
![image](https://user-images.githubusercontent.com/6665872/132492187-8345378a-91bb-41f7-bafd-42de0bf72dfd.png)

A long list of itens is not ideal but it can be desirable in the context of trainings linked to academic courses where a one wishes to preserve direct link between the academic learning outputs an contents, and the description the certificates will carry.

Maybe allowing the use of some HTML tags like `<br>` could resolve this?

jvdkwast:
Instead of adding line breaks we suggest to limit the amount of characters in the text field. This is to avoid that the certificate gets too many lines with competencies. We rather see a one line list and more info goes on the course description. 120 characters would fit nicely on 1 line.

PR #1475, opened 2025-01-20
Limit trained_competence to 120 characters

Xpirix:
Fix for #1366 

I think it's better to limit the character from the form instead of adjusting the max_length directly in the model to avoid changing the existing data.

For existing courses with `Trained competence` that exceed the limit of 120 characters, an error will be raised when updating the form:

image

I've also directly truncated this attribute in the certificate to avoid issues like #1366 for the existing course, so the user will also see that it doesn't fit and needs to be updated.


timlinux:
LGTM thanks!

dimasciput (approved):

Issue #1453, opened 2023-12-18
App allows to generate certificates without tokens


antoniolocandro:
I currently have no tokens in my organization yet I was able to generate the certificates, I see how this can be exploited by some organizations that may charge for the certificate but not pay for the tokens. The app should only allow to create certificates if tokens are available 
![image](https://github.com/kartoza/prj.app/assets/4627073/9bd00f99-da23-45fe-9ee3-32883d5036eb)

jvdkwast:
This is still the case.
![image](https://github.com/user-attachments/assets/19a24635-e289-4dcd-9eef-f6a9904e05b4)

Issue #1470, opened 2024-11-04
Generate all button generates all certificates even when credits are zero


jvdkwast:
I accidentally clicked the "Generate all certificates for this course" button instead of the "Download all issued certificates from this course" button. I just wanted to download the certificates of 4 participants that successfully completed the course.
![afbeelding](https://github.com/user-attachments/assets/34c6f21b-3802-45dd-8825-a873b6314090)

Although my credits are 0, it generated all certificates and I was not able to undo the action. Deleting a participant that shouldn't receive a certificate didn't help. The link to the issued certificate remains on the server.

My suggestions for a solution:

- "Generate all certificates for this course" should check if there are sufficient credits and shouldn't work if credits are insufficient.
- "Generate all certificates for this course" should give a warning be sure that the user really wants to generate all certificates.
- Offer an option to retract an issued certificate.

I also wonder that if I topup now, the credits will be used to pay for the unwanted issued certificates?

Xpirix:
Thank you for reporting this @jvdkwast 

Some parts of this issue should be fixed by #1478 .

> "Generate all certificates for this course" should give a warning be sure that the user really wants to generate all certificates.

Additionally, we should also be able to choose the list of attendees on that warning page.

> Offer an option to retract an issued certificate.

Is this something similar to the revoke certificate features?

> I also wonder that if I topup now, the credits will be used to pay for the unwanted issued certificates?

Yes, indeed. I think the best approach would be to choose the list of attendees on the warning page.

PR #1478, opened 2025-01-22
Enforce credits checks

Xpirix:
Fix for #1453 

- Disable the `Generate all certificates button` when there's no available credit
- Enforce the verification of paid certificates

image

### DIFF
--- a/django_project/certification/templates/course/detail.html
+++ b/django_project/certification/templates/course/detail.html
@@ -174,12 +174,12 @@
             data-title="Download all issued certificates from this course">
             <span class="glyphicon glyphicon-download-alt"></span>
             </a>
-        {% if course.certifying_organisation.organisation_credits > attendees.count and attendees.count > 0 %}
+        {% if course.certifying_organisation.organisation_credits >= attendees.count and attendees.count > 0 %}
             <a class="btn btn-default btn-sm tooltip-toggle btn-info-focus"
             href='{% url "generate-all-certificate" project_slug=course.certifying_organisation.project.slug organisation_slug=course.certifying_organisation.slug course_slug=course.slug %}'
                 data-title="Generate all certificate for this course"
             >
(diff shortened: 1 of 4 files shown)


django_project/certification/templates/course/detail.html
```diff
@@ -174,11 +174,17 @@ <h4 class="panel-heading" style="padding-left:10px; padding-right: 10px">
             data-title="Download all issued certificates from this course">
             <span class="glyphicon glyphicon-download-alt"></span>
             </a>
-        <a class="btn btn-default btn-sm tooltip-toggle btn-info-focus"
-           href='{% url "generate-all-certificate" project_slug=course.certifying_organisation.project.slug organisation_slug=course.certifying_organisation.slug course_slug=course.slug %}'
-            data-title="Generate all certificate for this course">
+        {% if course.certifying_organisation.organisation_credits > 0 %}
```
timlinux: What if it is >0 but not enough for all?

timlinux:
LGTM once my comment is addressed, thanks!

django_project/certification/templates/course/detail.html
Xpirix: @timlinux We can still generate certificates but only the amount of attendees corresponding to the credit will get theirs generated. The others will have to wait for the credit top-up to be able to get the certificate.

image

Xpirix: @timlinux I have updated the code to also take into consideration #1470. The button will be disabled if the credits are not sufficient to generate all certificates for all attendees.

image

Xpirix:
Updates according to #1470:

-  Disable the "Generate all certificates for this course"  if credits are not sufficient.
- Show a warning page when pushing the "Generate all certificates for this course" button

https://github.com/user-attachments/assets/3275aae4-0bc9-49cf-b567-18a1867ec2f7

dimasciput (approved):

PR #1480, opened 2025-01-22
Use the same time limit for editing attendee's and revoking certificate

### LINKED ISSUE #1443: Edit the name of the attendees (related to #1435)
We recently issued a batch of certificates for a course that finished in October 2022 but whose final assessment (and therefore the right to a QGIS certification on top of the academic credits) finished in March.

We have a few instances where the name on the certificate needs to be corrected, but we cannot change it now because you can only do it seven days after the end of the course.

**These seven days should be counted after the EMISSION of the certificate** and not the date inscribed in the certificate. Now, as it stands, the only option is to re-issue the certificates using new credits, which is not optimal.  

Is there a way to fix this issue while applying a more permanent fix?

Thanks

Xpirix:
Additional fix for https://github.com/kartoza/prj.app/issues/1435

Should also fix #1443

### DIFF
--- a/django_project/certification/templates/course/detail.html
+++ b/django_project/certification/templates/course/detail.html
@@ -244,7 +244,7 @@
                         {% endif %}
                         {% if attendee.attendee.pk in certificates %}
                             {% if attendee.attendee.pk in paid_certificates %}
-                                {% if course.editable %}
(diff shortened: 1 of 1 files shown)


timlinux:
LGTM

Issue #1390, opened 2022-01-07
Auto-reject Pending certifying organisations


sumandari:
We should auto-reject orgs after a configurable time frame e.g. 6 months

jvdkwast:
Kurt and I suggest to expand it to 1 year.

PR #1481, opened 2025-01-22
Auto reject organisation

Xpirix:
Fix for #1390 

- Added a command that rejects pending organisations older than one year
- Need to add the following cronjob: `0 0 * * * cd ~/prj.app/deployment && docker compose exec devweb python manage.py reject_pending_organisations`

I think it would be better to use celery later when we apply the stack upgrade. For now, it's more suitable to set it as a cronjob.


timlinux:
LGTM

Issue #1411, opened 2022-03-25
QGIS Changelog list only shows partial list


timlinux:
Can we make the list view of changelogs show the complete list please?

![image](https://user-images.githubusercontent.com/178003/160080015-f63a22d7-05a2-437f-b6d9-266965364050.png)

Should go all the way back to v2

PR #1483, opened 2025-01-29
List all existing versions

Xpirix:
Fix for #1411 

image

image


timlinux:
Better fix is to back fill the release dates and make the release date mandatory.

Issue #1456, opened 2023-12-19
Spinner shouldn't allow negative values


antoniolocandro:
I know the stripe button is not available but still not a good UI/UX

![image](https://github.com/kartoza/prj.app/assets/4627073/47ff4d69-643b-48b6-ab11-ab7ebd74dddc)

PR #1484, opened 2025-01-29
Only allow positive numbers in the credits input

Xpirix:
Fix for #1456 

- Set the minimum allowed value to 1
- Only numbers are allowed to be typed in the input, typing any other characters won't have any effect.

https://github.com/user-attachments/assets/cca5f0d7-4393-453d-873a-e19eb8b97a11


timlinux:
LGTM Thanks

dimasciput (approved):

